PR #14, opened 2020-07-03
service: add records service

ppanero:
Decision taken based on https://codimd.web.cern.ch/dX2AYr7KRLGZrtqY3-G6iA

### DIFF
--- a/invenio_resources/service/actions.py
+++ b/invenio_resources/service/actions.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+#
+# Invenio-Records-Agent is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Actions API."""
+
+
+class Action:
+    """Action interface."""
+
+    def __init__(self, state):
+        """Initialize action."""
+        self.state = state
+
+    def run(self):
+        """Run action."""
+        raise NotImplementedError()
+
+
+class MacroAction(Action):
+    """Macro action for running multiple actions."""
+
+    #: List of actions to execute
+    actions = []
+
+    def iter_actions(self):
+        """Iterate over actions."""
+        for a in self.actions:
+            yield a(self.state)
+
+    def run(self):
+        """Run each action defined."""
+        for a in self.iter_actions():
+            a.run()
+
+
+class PublishAction(MacroAction):
+    """Record publishing action."""
+
+    actions = []
--- a/invenio_resources/service/search/engine.py
+++ b/invenio_resources/service/search/engine.py
@@ -1,0 +1,47 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+#
+# Invenio-Resources is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Search engine."""
+
+from .query import QueryInterpreter
+
+
+class SearchEngine():
+    """Search records with specific configuration."""
+
+    def __init__(self, search_cls, query_interpreter=QueryInterpreter(),
+                 *args, **kwargs):
+        """Constructor."""
+        self.search = search_cls(*args, **kwargs)
+        self.query_interpreter = query_interpreter
+
+    def search_arguments(self, pagination=None, preference=True, version=True,
+                         extras=None):
+        """Adds arguments to the search object."""
+        # Avoid query bounce problem
+        if preference:
+            self.search = self.search.with_preference_param()
+        # Add document version to ES response
+        if version:
+            self.search = self.search.params(version=True)
+        # Add other aguments
+        if pagination:
+            self.search = self.search[
+                pagination["from_idx"]: pagination["to_idx"]
+            ]
+
+        # Add extra args
+        if extras:
+            self.search = self.search.extra(**extras)
+
+    def parse_query(self, querystring):
+        """Parses the query based on the interpreter."""
+        return self.query_interpreter.parse(querystring)
+
+    def execute_search(self, query):
+        """Executes an already parsed query."""
+        return self.search.query(query).execute()
--- a/invenio_resources/service/search/serializers.py
+++ b/invenio_resources/service/search/serializers.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+#
+# Invenio-Resources is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Invenio Resources module to create REST APIs."""
+
+from dateutil import parser
+
+
+def es_to_record(record_hit, record_cls):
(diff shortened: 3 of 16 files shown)


tests/service/conftest.py
```diff
@@ -0,0 +1,86 @@
...
+@pytest.fixture()
+def fake_record_db():
+    """A fake record."""
+    return {
+        uuid4(): {'title': 'Fake record'},
+    }
+
+
+@pytest.fixture()
+def fake_record_cls(fake_record_db):
```
fenekku, line 44: That's an interesting pattern!

tests/service/test_service.py
```diff
@@ -0,0 +1,29 @@
...
+# details.
+
+"""Agents tests."""
+
+from uuid import uuid4
+
+import pytest
+from flask_principal import Identity, UserNeed
+
+from invenio_resources.service import RecordService, RecordServiceFactory
```
fenekku, line 16: minor: remove all

tests/service/test_service.py
```diff
@@ -0,0 +1,29 @@
...
+    """Get a record."""
+    for recid, fake_record in fake_record_db.items():
+        recstate = service_cls.get(recid, identity=identity_simple)
+        assert recstate.record == fake_record
+        assert recstate.id == str(recid)
+
+    # test not found
+    # test not found
+    # test pid resolver errors
+    # test permission errors
```
fenekku, line 29: Shelved: These additional tests are crucial. A lot of machinery is put into place in the code of this commit, but without the tests, it is hard to know why or if the machinery works well for the use cases we have (which would be illustrated by the tests). I am sure the tests are coming in next PR ;)

invenio_resources/service/state.py
```diff
@@ -0,0 +1,30 @@
...
+# Copyright (C) 2020 CERN.
+#
+# Invenio-Records-Agent is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""State API."""
+
+
+class RecordState:
```
fenekku, line 12: Medium: `State` is probably too generic for us (every object has state, "state" could be used for anything...). My suggestion for the name: `IdentifiedRecord` . It represents a Record and its identifiers. It has an id and the PersistentIdentifiers mapping to it. It is identified.

This is a really good object as it solves our [data clump](https://en.wikipedia.org/wiki/Data_clump) problem. It should almost be passed around in most of Invenio IMHO.

invenio_resources/service/state.py
```diff
@@ -0,0 +1,30 @@
...
+        self.id = pid.pid_value
+        self.pids = [pid]
+        self.record = record
+
+    def is_revision(self, revision_id):
+        """Check if record is in a specific revision."""
+        return str(self.record.revision_id) == str(revision_id)
+
+
+class TombstoneState(RecordState):
```
fenekku, line 26: minor: eeeeh, would probably use `is_tombstone` on `RecordState` instead for now :)

invenio_resources/service/service.py
```diff
@@ -0,0 +1,189 @@
...
+    #
+    @classmethod
+    def get(cls, id_, identity):
+        """Retrieve a record."""
+        pid, record = cls.resolve(id_)
+        cls.require_permission(identity, "read", record=record)
+        # Todo: how do we deal with tombstone pages
+        return cls.factory.record_state(pid=pid, record=record)
+
+    @classmethod
```
fenekku, line 133: I am ignoring below from review because tests are pending for these use cases.

invenio_resources/service/service.py
```diff
@@ -0,0 +1,189 @@
...
+        """Resolve a persistent identifier to a record."""
+        return cls.factory.resolver().resolve(id_)
+
+    #
+    # High-level API
+    #
+    @classmethod
+    def get(cls, id_, identity):
+        """Retrieve a record."""
+        pid, record = cls.resolve(id_)
```
fenekku, line 128: Normal: we could have the `resolve` method just return the `State` (`IdentifiedRecord` plug-plug :wink: ). `cls` has everything needed.

invenio_resources/service/service.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def record_state(self, **kwargs):
+        """Create a new item state."""
+        return self._config.record_state_cls(**kwargs)
+
+
+class RecordService:
+    """Record Service interface."""
+
+    factory = RecordServiceFactory(RecordServiceConfig)
```
fenekku, line 103: Would rename `RecordServiceConfig` to `RecordServiceFactoryConfig` since it is the factory's config. (feels like déjà-vu). `RecordServiceConfig` reads as the configuration for `RecordService` which it turns out not to be.

invenio_resources/service/service.py
```diff
@@ -0,0 +1,189 @@
...
+
+    #
+    # Permissions checking
+    #
+    @classmethod
+    def require_permission(cls, identity, action_name, **kwargs):
+        """Require a specific permission from the permission policy."""
+        if not cls.factory.permission(action_name, **kwargs).allows(identity):
+            raise PermissionDeniedError(action_name)
```
fenekku, line 112: TODO: integration with invenio-records-permission

invenio_resources/service/search/serializers.py
```diff
@@ -0,0 +1,37 @@
...
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Invenio Resources module to create REST APIs."""
+
+from dateutil import parser
+
+
+def es_to_record(record_hit, record_cls):
+    """Transforms an Elasticsearch record hit into a Record."""
+    _id = record_hit["_id"]
```
fenekku, line 15: Let's do a deep copy of `record_hit` first here. Otherwise we are mutating `record_hit` which will lead to strange behaviour.

invenio_resources/service/search/engine.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+#
+# Invenio-Resources is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Search engine."""
```
fenekku, line 8: Delay review until search tests are in.

invenio_resources/service/actions.py
```diff
@@ -0,0 +1,44 @@
...
+    def run(self):
+        """Run each action defined."""
+        for a in self.iter_actions():
+            a.run()
+
+
+class PublishAction(MacroAction):
+    """Record publishing action."""
+
+    actions = []
```
fenekku, line 44: Looks like a command pattern for record operation. Will be clearer when tests are in.

invenio_resources/service/state.py
ppanero: Tombstone page is not implemented, it was there as an idea. It might be that is not even a state but a mere exception handled different (as in invenio-records-rest atm). Would shelve.

invenio_resources/service/service.py
ppanero: Encapsulation sounds good to me, as long as is done from `resolve` method here so we dont have to touch resolver in pidstore.

PR #15, opened 2020-07-03
global: implement record deletion

ppanero:
Tests are failing because the view is trying to give a full response when the `delete` method response body is empty. Needs to be addressed in flask-resources.


ppanero:
Closed in favour of #18

PR #21, opened 2020-07-09
service: support data validation

ppanero:
- Support using marshmallow schemas
- Refactor current schemas to support a non RDM tight metadata schema
- Refactor tests to reflect the new schemas
- Tests:
    - Test data validator interface (for coverage reasons only)
    - Tests default marshmallow data validator
    - Test  marshmallow data validator with custom schema
    - Test request with validation error commented out, because it throws an exception. Error handling will solve this issue.

Careful if https://github.com/inveniosoftware/invenio-resources/pull/24 is merged before. It implies a rebase and refactor on the tests structure.

### DIFF
--- a/invenio_records_resources/resources/record.py
+++ b/invenio_records_resources/resources/record.py
@@ -51,8 +51,7 @@ class RecordResource(CollectionResource):
     #
     def search(self, *args, **kwargs):
         """Perform a search over the items."""
-        # TODO fix identity extraction
-        identity = g.identiy
+        identity = g.identity
         record_search = self.service_cls.search(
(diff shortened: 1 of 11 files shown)


invenio_records_resources/resources/record.py
```diff
@@ -51,7 +51,6 @@ def __init__(
     #
     def search(self, *args, **kwargs):
         """Perform a search over the items."""
-        # TODO fix identity extraction
         identity = g.identiy
```
zzacharo, line 54: @ppanero Sorry there is typo :(

PR #25, opened 2020-07-09
tests: make them go

fenekku:
Subsequent API stubs will come later. It took most of my time to discover that these tests were broken and it wasn't my machine.


fenekku:
@ppanero You are right that comments are not the way to go. What about using `@pytest.mark.skip`? As in 

```python
@pytest.mark.skip(reason="TODO: proper ES index cleanup needed before can be tested")
def test_create_search_record():
```

This would make it clearer.

PR #28, opened 2020-07-13
errors: configure resource with error mapping

zzacharo:
Depends on https://github.com/inveniosoftware/flask-resources/pull/67
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/214

### DIFF
--- a/invenio_records_resources/resources/record.py
+++ b/invenio_records_resources/resources/record.py
@@ -12,13 +12,16 @@
 from flask import g
 from flask_resources import CollectionResource
 from flask_resources.context import resource_requestctx
+from flask_resources.errors import HTTPJSONException, create_errormap_handler
 from flask_resources.loaders import RequestLoader
 from flask_resources.resources import ResourceConfig
+from invenio_pidstore.errors import PIDDeletedError
 
 from ..responses import RecordResponse
 from ..schemas import RecordSchemaJSONV1
 from ..serializers import RecordJSONSerializer
 from ..services import RecordService
+from ..services.errors import InvalidQueryError, PermissionDeniedError
 
 
 class RecordResourceConfig(ResourceConfig):
@@ -30,6 +33,26 @@ class RecordResourceConfig(ResourceConfig):
         "application/json": RecordResponse(
             RecordJSONSerializer(schema=RecordSchemaJSONV1)
         )
+    }
+    error_map = {
+        InvalidQueryError: create_errormap_handler(
+            HTTPJSONException(
+                code=400,
+                description="Invalid query syntax.",
+            )
+        ),
+        PermissionDeniedError: create_errormap_handler(
+            HTTPJSONException(
+                code=403,
+                description="Permission denied.",
+            )
+        ),
+        PIDDeletedError: create_errormap_handler(
+            HTTPJSONException(
+                code=410,
+                description="The record has been deleted.",
+            )
+        ),
     }
 
 
@@ -54,7 +77,6 @@ class RecordResource(CollectionResource):
             identity=identity,
             pagination=resource_requestctx.request_args.get("pagination"),
         )
-
         return record_search, 200
 
     def create(self, *args, **kwargs):
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,7 @@ install_requires = [
     "invenio-indexer>=1.1.1",
     "invenio-records>=1.3.2",
     "invenio-rest>=1.2.1",
-    # "flask-resources", # FIXME: Currently in dev
+    "flask-resources>=0.2.0,<1.0.0",
     # Service
     "invenio-accounts>=1.3.0",
     "invenio-files-rest>=1.2.0",
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -16,6 +16,7 @@ from datetime import date
 
 import pytest
 from flask import Flask
+from flask_resources.errors import handle_http_exception
 from flask_security.utils import hash_password
 from invenio_access.models import ActionUsers
 from invenio_access.permissions import superuser_access
@@ -24,6 +25,7 @@ from invenio_accounts.models import Role
 from invenio_app.factory import create_api as _create_api
 from invenio_records_permissions.generators import Admin, AnyUser
 from invenio_records_permissions.policies.records import RecordPermissionPolicy
+from werkzeug.exceptions import HTTPException
 
 from invenio_records_resources.resources import RecordResource, \
     RecordResourceConfig
@@ -117,5 +119,6 @@ def app(app):
         RecordResource(service=RecordService()).as_blueprint("base_resource")
     )
     app.register_blueprint(custom_bp)
+    app.register_error_handler(HTTPException, handle_http_exception)
     with app.app_context():
         yield app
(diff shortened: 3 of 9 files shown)


invenio_records_resources/resources/record.py
```diff
@@ -51,15 +74,16 @@ def __init__(
     #
     def search(self, *args, **kwargs):
         """Perform a search over the items."""
+        # TODO fix identity extraction
```
ppanero, line 77: major: old comment.

invenio_records_resources/resources/record.py
```diff
@@ -51,15 +74,16 @@ def __init__(
...
         identity = g.identity
-        record_search = self.service_cls.search(
-            querystring=resource_requestctx.request_args.get("q", ""),
-            identity=identity,
-            pagination=resource_requestctx.request_args.get("pagination"),
+        return (
```
ppanero, line 79: Minor: I'm fine with this change. But it feels changing code that is totally equivalent to the previous for a mere "programming" preference. Might affect history.

invenio_records_resources/resources/record.py
```diff
@@ -71,7 +95,8 @@ def read(self, *args, **kwargs):
         identity = g.identity
         return (
             self.service_cls.get(
-                id_=resource_requestctx.route["pid_value"], identity=identity
+                id_=resource_requestctx.route["pid_value"],
```
ppanero, line 98: Minor: Same as above

invenio_records_resources/service/service.py
```diff
@@ -206,6 +206,7 @@ def delete(cls, id_, identity):
 
         record.delete()
         # TODO: mark all PIDs as DELETED
```
ppanero, line 208: moderate: Remove `todo`. It got fixed on the added line below no?

tests/test_record_resource.py
```diff
@@ -35,7 +35,6 @@ def test_create_read_record(client, input_record):
```
ppanero, line 35: Moderate (Applies to the whole file):

Align with @fenekku on this:

- There is a refactor
- @fenekku 's PR uses `/records` (I would say this is the way to go)
- This PR has a different way of handling the skip (IMO it's better since tests will still run on travis, but we avoid the hassle in local).

invenio_records_resources/resources/record.py
zzacharo: I did it just to normalize all the views as we had 2 ways of returning the view's result :) It's one less line though...But I am fine to change it back if you see it as problematic!

tests/test_record_resource.py
zzacharo: Yes I saw that :/ If we merge his PR first I will do the changes :) It's just that currently I need to keep it like that! But, I totally agree with the changes he made!

zzacharo: I will also need to apply the error handling the other views he created most probably, so it might be better to merge his first :)

invenio_records_resources/resources/record.py
```diff
@@ -46,22 +69,21 @@ def __init__(
     ):
         """Constructor."""
         super(RecordResource, self).__init__(*args, **kwargs)
-        self.service_cls = service_cls
```
fenekku, line 49: Moderate: Was this supposed to be removed? It's being used below...

invenio_records_resources/resources/record.py
fenekku: I am with Pablo on this one :)

invenio_records_resources/service/search/query.py
```diff
@@ -26,13 +29,13 @@ def __init__(self, query_parser=None, **kwargs):
         self.query_parser = query_parser or _default_parser
         self.params = kwargs
 
-    def parse(self, queryparser):
+    def parse(self, querystring):
```
fenekku, line 32: Hmm the `querystring` term is a little overloaded: https://en.wikipedia.org/wiki/Query_string .. It means the whole part of the URL after the `?` e.g., for `https://example.com/?q=tag:good+tag:alright&lang=fr` it's `q=tag:good+tag:alright&lang=fr`, but in the code here querystring is referring to `tag:good+tag:alright` right? What if we used the term `searchstring` to resolve this ambiguity? (other names are possible too).

tests/test_permissions.py
```diff
@@ -55,20 +55,22 @@ def test_create_record_permissions(app_with_custom_permissions, client,
     user1 = users['user1']
     user2 = users['user2']
     # Create new record as anonymous
-    with pytest.raises(PermissionDeniedError) as e:
-        response = client.post(
-            "/records", headers=HEADERS, data=json.dumps(input_record)
-        )
+    response = client.post(
+        "/records", headers=HEADERS, data=json.dumps(input_record)
+    )
```
fenekku, line 60: :100:

tests/test_record_resource.py
```diff
@@ -52,7 +52,8 @@ def test_create_read_record(client, input_record):
 # because the CI creates a new container for each test run. It will work
 # once locally, but fail on subsequent run.
 # TODO: FIX to make it clean up itself. Adding es_clear didn't work
-@pytest.mark.skip(reason="Fix to make it clean up itself")
+@pytest.mark.skipif(
+    os.environ.get('CI') != 'true', reason="Fix to make it clean up itself")
```
fenekku, line 56: Ah very nice!

invenio_records_resources/resources/record.py
zzacharo: yes sorry I added it back :)

invenio_records_resources/resources/record.py
```diff
@@ -31,6 +34,26 @@ class RecordResourceConfig(ResourceConfig):
             RecordJSONSerializer(schema=RecordSchemaJSONV1)
         )
     }
+    error_map = {
+        InvalidQueryError: create_errormap_handler(
```
ppanero, line 38: moderate: is there a way of reusing the `description` of the base error. i.e. we are duplicating the string coming form other exceptions (e.g. `PermissionDeniedError`, etc.) would be nice to just reuse.

setup.py
```diff
@@ -61,7 +61,7 @@
     "invenio-indexer>=1.1.1",
     "invenio-records>=1.3.2",
     "invenio-rest>=1.2.1",
-    # "flask-resources", # FIXME: Currently in dev
+    "flask-resources~=0.2.0",
```
ppanero, line 64: Moderate: Since we are under heavy development. Shall we just lower pin until first? We ahve been doing that in invenio-rdm-records, invenio-app-rdm, etc. (https://github.com/inveniosoftware/invenio-rdm-records/blob/master/setup.py#L78)

```suggestion
    "flask-resources>=0.2.0,<1.0.0",
```

zzacharo: @ppanero We do test `lower_requirements` so a bump most of the time will do be needed. I am ok with both options :)

tests/conftest.py
```diff
@@ -117,5 +119,6 @@ def app(app):
         RecordResource(service=RecordService()).as_blueprint("base_resource")
     )
     app.register_blueprint(custom_bp)
+    app.register_error_handler(HTTPException, handle_http_exception)
```
fenekku, line 122: minor: like discussed this could be set on the blueprint. For tests this is fine.

PR #30, opened 2020-07-13
global: stub records API

### LINKED ISSUE #22: Implement stubs for REST API endpoints
In Invenio-Records-Resources and Invenio-Drafts-Resources.

- [x] ``/records``: GET, POST
- [x] ``/records/:id``: GET, DELETE
- [x] ``/records/:id/versions`` GET, POST
- [x] ``/records/:id/files``: GET
- [x] ``/records/:id/files/:key`` GET
- [x] ``/records/:id/files/:key/download`` GET
- [x] ``/user/records`` GET
- [x] ``/records/:id/draft``: GET, POST, PUT, DELETE
- [x] ``/records/:id/draft/actions/:action``: POST
- [x] ``/records/:id/draft/files`` GET, POST, PUT, DELETE
- [x] ``/records/:id/draft/files/:key``: GET, PUT, DELETE
- [x] ``/records/:id/draft/files/:key/download`` GET

fenekku:
closes #22 for real

(the skip is there but Zach will replace it)

### DIFF
--- a/invenio_records_resources/resources/file.py
+++ b/invenio_records_resources/resources/file.py
@@ -1,0 +1,56 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+# Copyright (C) 2020 Northwestern University.
+#
+# Invenio-Records-Resources is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Invenio Record File Resources."""
+
+from flask_resources import CollectionResource, SingletonResource
+from flask_resources.context import resource_requestctx
+from flask_resources.resources import ResourceConfig
+
+
+class FileResourceConfig(ResourceConfig):
+    """Record resource config."""
+
+    item_route = "/records/<pid_value>/files/<key>"
+    list_route = "/records/<pid_value>/files"
+
+
+class FileResource(CollectionResource):
+    """File resource."""
+
+    default_config = FileResourceConfig
+
+    def search(self, *args, **kwargs):
+        """List items."""
+        return {"TODO": "IMPLEMENT ME"}, 200
+
+    def read(self, *args, **kwargs):
+        """Read an item."""
+        return {"TODO": "IMPLEMENT ME"}, 200
+
+
+class FileActionResourceConfig(ResourceConfig):
+    """Record resource config."""
+
+    # QUESTIONs:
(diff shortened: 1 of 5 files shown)


invenio_records_resources/resources/file.py
```diff
@@ -0,0 +1,56 @@
...
+    def read(self, *args, **kwargs):
+        """Read an item."""
+        return {"TODO": "IMPLEMENT ME"}, 200
+
+
+class FileActionResourceConfig(ResourceConfig):
+    """Record resource config."""
+
+    # QUESTIONs:
+    # 1- Shouldn't the item_route be used for SingletonResource actually?
```
ppanero, line 42: This is a good question. Maybe we can have short discussion after standup. Im in doubts too.

invenio_records_resources/resources/record.py
```diff
@@ -35,15 +36,16 @@ class RecordResourceConfig(ResourceConfig):
...
+
     def __init__(
         self,
-        config=RecordResourceConfig(),
         service_cls=RecordService,
         *args,
         **kwargs
     ):
         """Constructor."""
-        super(RecordResource, self).__init__(config=config, *args, **kwargs)
```
ppanero, line 46: Shouldn't this be something like:

```python
config = config if config else default_config
```

Otherwise where si the logic for "default to `default_config` if no config is passed`. I know it was somewhere, just cannot find it now.

tests/test_record_resource.py
```diff
@@ -52,46 +52,47 @@ def test_create_read_record(client, input_record):
 # because the CI creates a new container for each test run. It will work
 # once locally, but fail on subsequent run.
 # TODO: FIX to make it clean up itself. Adding es_clear didn't work
-# def test_create_search_record(client, input_record):
-#     """Test record search."""
-#     # Search records, should return empty
-#     response = client.get("/records_v2", headers=HEADERS)
-#     assert response.status_code == 200
+@pytest.mark.skip(reason="Fix to make it clean up itself")
```
ppanero, line 55: @zzacharo's PR has a different approach. Be careful in merge order according to what you decide should prevail.

fenekku: Zach's approach should prevail

invenio_records_resources/resources/record.py
fenekku: Ah you mean this:  https://github.com/inveniosoftware/flask-resources/blob/master/flask_resources/resources.py#L45 right?

Issue #9, opened 2020-06-18
serializers: common interface to serialize drafts, records, tombstone pages, etc.


ppanero:
Example of implementation in `flask-resources`, it was removed from there because no default class needed to implement it.

```
class SerializableMixin:
    """Serializable Interface.

    Objects that will be serialized must implement it.
    """

    @property
    def object():
        """Returns the object itself."""
        raise NotImplementedError()

    @property
    def id():
        """Returns the object id."""
        raise NotImplementedError()

    @property
    def version():
        """Returns the object version or revision."""
        raise NotImplementedError()

    @property
    def last_modified():
        """Returns the date of the last modification."""
        raise NotImplementedError()
```

PR #31, opened 2020-07-13
26 datacite xml

fenekku:
This is more of a proof of concept than an actual serializer that should be used.

closes: #26

### DIFF
--- a/invenio_records_resources/serializers/xml.py
+++ b/invenio_records_resources/serializers/xml.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+# Copyright (C) 2020 Northwestern University.
+#
+# Invenio-Records-Resources is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Invenio Resources basic XML serialization module."""
+
+import xmltodict
+from flask_resources.serializers import SerializerMixin
+
+from .json import RecordJSONSerializer
+
+
+class RecordXMLSerializer(SerializerMixin):
+    """Record JSON serializer implementation."""
+
+    def serialize_object(self, obj, response_ctx=None, *args, **kwargs):
+        """Dump the object into an XML string."""
+        if obj:  # e.g. delete op has no return body
+            json_serializer = RecordJSONSerializer()
+            processed_record = json_serializer._process_record(obj)
+            return xmltodict.unparse({"root": processed_record})
+        else:
+            return ""
+
+    def serialize_object_list(
+        self, obj_list, response_ctx=None, *args, **kwargs
+    ):
+        """Dump the object list into an XML string.
+
+        :param: obj_list a RecordSearchState object
+        """
+        raise NotImplementedError()
--- a/setup.py
+++ b/setup.py
@@ -66,6 +66,7 @@ install_requires = [
     "invenio-accounts>=1.3.0",
     "invenio-files-rest>=1.2.0",
     "invenio-records-permissions>=0.8.0",
+    "xmltodict~=0.12.0",
 ]
 
 packages = find_packages()
--- a/invenio_records_resources/resources/file.py
+++ b/invenio_records_resources/resources/file.py
@@ -1,0 +1,56 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+# Copyright (C) 2020 Northwestern University.
+#
+# Invenio-Records-Resources is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Invenio Record File Resources."""
+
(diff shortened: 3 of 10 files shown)


invenio_records_resources/serializers/xml.py
```diff
@@ -0,0 +1,37 @@
...
+
+class RecordXMLSerializer(SerializerMixin):
+    """Record JSON serializer implementation."""
+
+    def serialize_object(self, obj, response_ctx=None, *args, **kwargs):
+        """Dump the object into an XML string."""
+        if obj:  # e.g. delete op has no return body
+            json_serializer = RecordJSONSerializer()
+            processed_record = json_serializer._process_record(obj)
+            return xmltodict.unparse({"root": processed_record})
```
ppanero, line 26: Moderate/Major: I understand this is for "demo" purposes. However, dumping json here makes the whole object end up in a single key - `root`. IMHO, each key should be its own XML one.

```python
>>> a = {"test": 1, "another": 2}
>>> xmltodict.unparse({"root":a})
'<?xml version="1.0" encoding="utf-8"?>\n<root><test>1</test><another>2</another></root>'
>>> xmltodict.unparse({"root":json.dumps(a)})
'<?xml version="1.0" encoding="utf-8"?>\n<root>{"test": 1, "another": 2}</root>'
```

```suggestion
            return xmltodict.unparse({"root": obj})
```

I think this is not noticed because the integration tests do not check the contents of the `metadata` field, which is where the output of this function ends up.

extra karma: removes the dependency form `JSONSerializer` 😇

invenio_records_resources/serializers/xml.py
```diff
@@ -0,0 +1,37 @@
...
+        else:
+            return ""
+
+    def serialize_object_list(
+        self, obj_list, response_ctx=None, *args, **kwargs
+    ):
+        """Dump the object list into an XML string.
```
ppanero, line 33: Minor: Don't we want a full example? I think the idea is to have a fully functional one. Not just for demo. @lnielsen can confirm what was the plan :)

setup.py
```diff
@@ -66,6 +66,7 @@
     "invenio-accounts>=1.3.0",
     "invenio-files-rest>=1.2.0",
     "invenio-records-permissions>=0.8.0",
+    "xmltodict~=0.12.0",
```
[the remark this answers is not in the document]
slint: We can "pilot" using tilda `~` requirements, and review on a per-package basis what's the best practice (e.g. in this case we would trust `xmltodict 0.12.*`)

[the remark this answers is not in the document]
fenekku: :+1: 

To be clear (and I ranted about this before :sweat_smile:), I am in favour of putting lids on dependency versions now. Otherwise we have to constantly react to upstreams breaking.

invenio_records_resources/serializers/xml.py
fenekku: Oh, this is subtle :neckbeard: :

```python
xmltodict.unparse({"root": json.dumps(a)})
'<?xml version="1.0" encoding="utf-8"?>\n<root>{"test": 1, "another": 2}</root>'
```

is because `json.dumps(a)` returns a string. However, `json_serializer._process_record(obj)` returns a `dict` and that makes the whole difference. So we are good here! (the tests would have broken because we are iterating on the root and getting xml `Element` objects that respond to the `tag()` message. `dict` keys don't. ;) )

invenio_records_resources/serializers/xml.py
fenekku: Definitely would like to hear @lnielsen take because the XML serializer isn't really real: it can be used for Datacite XML or for DublinCore XML and those need to control top-level xml which would be better served by a DataciteXMLSerializer or DublinCoreXMLSerialzer that manipulates those.

invenio_records_resources/serializers/xml.py
ppanero: Hmm... you are right. Sorry I missunderstood that. Still I wouldn't make one serializer dependent on the other. We can make an independent method out of `_process_record`? It does actually not relate to any `self`.

invenio_records_resources/serializers/xml.py
ppanero: Hmm I see your point. However, the JSONSerializer is not "real" either. Is plain JSON, is not DataCiteJSON (that comes from RDM-Records datamodel... same goes for XML).

invenio_records_resources/serializers/xml.py
fenekku: I think extracting _process_record makes sense, but I would wait a little bit to determine where to extract it. I think the works we are both doing will help us see that. For now, I would say let's go with that.

invenio_records_resources/serializers/xml.py
fenekku: Let's shelve this or discuss it at the standup. (This issue, is perhaps getting at the core of it: https://github.com/inveniosoftware/invenio-records-resources/issues/9 better?)

PR #32, opened 2020-07-14
23 records + files service stubs

### LINKED ISSUE #23: Implement stubs for services (Records, Files)

fenekku:
closes #23 

Issues for [extracting link generation](https://github.com/inveniosoftware/invenio-records-resources/issues/33) and for finding a good pattern to [abstract Records coming from DB and from ES](https://github.com/inveniosoftware/invenio-records-resources/issues/34) have been created.

**EDIT**
*The below was discussed in person.*

This is one part of the #23 service stubs work: specifically for Record and Files in invenio-records-resources (other service stubs will be in invenio-drafts-resources). The goal of the work was to surface trickiness / patterns / names, so even if this needs to be rebased and all that, there are a bunch of interesting outcomes:

- `RecordState` naming: we were wandering what to name this thing in Service / Resource. The way I see it is through the real-time-strategy game analogy :video_game: : you have a resource Lumber or Space Mineral, but you transact in units of that resource: units of lumber or units of minerals. So... I went with `resource_unit` and `resource_unit_cls`. Now in its own context, this unit of resource may have an independent name. For example: `RecordState` -> `IdentifiedRecord`. `IdentifiedRecord` is assigned to the `resource_unit_cls` configuration, but we don't use "ResourceUnit" in its name because that thing is used in other contexts than just Resource and Service.  (Note that I haven't done the RecordState renaming mentioned yet).
- Invenio Records are pretty "generic" (bibliographic record, authors can all fit). In fact, using RecordService for all these cases makes perfect sense. Only the configuration needs to change. This is why this PR refactors RecordService methods to be instance methods, so that the same RecordService can be used with different configurations. @ppanero in particular we *may* get enough mileage from RecordService to not even need a DraftService ('cause a draft is just a record). To be verified for sure.
- Even though, I split RecordService and Service and made RecordService inherit from it, the split of methods still needs refinement (pretty sure it's not quite there yet). It is unclear if the configuration for Service as it is in this PR is applicable to File(Metadata)Service... The resource unit of the File related services is not a Record and so the configuration related methods we have so far may be too specialized to even be in Service.
- Changed `get` to `read` in Service to completely parallel Resource. No reason not to do so and just introduces a way to make a mistake.
- When it comes to the return value of `search`, decreeing that it must be an iterable of `resource_unit` is an idea... but it's incomplete: we probably want an object that carries an iterable of `resource_unit`s + some top-level data (pagination and the such). So it might need to be configured with `resource_list_cls` on the configuration.   

Thoughts?

Next up would be looking at patterns for representing a record that is backed by the DB and one that is backed by ES equivalently from an interface point of view. Need to read-up on some patterns...


ppanero:
>  @ppanero in particular we may get enough mileage from RecordService to not even need a DraftService ('cause a draft is just a record). To be verified for sure.

It might be, some things need to be overwritten (e.g. draft create would make use of the "record" service that it is drafting in order to create it). But yes, I agree. in #invenio-drafts-resources/1 I was on the way to refactor a bit for this.

> When it comes to the return value of search, decreeing that it must be an iterable of resource_unit is an idea... but it's incomplete: we probably want an object that carries an iterable of resource_units + some top-level data (pagination and the such). So it might need to be configured with resource_list_cls on the configuration.

Fully agree, I had implemented the `RecordSearchState` which is a first step towards it. Not complete in any case though.

> Next up would be looking at patterns for representing a record that is backed by the DB and one that is backed by ES equivalently from an interface point of view. Need to read-up on some patterns...

I'm quite interested in this topic :) Eager to see what comes out from it.

The rest of points --> Nothing else to add --> ✅ 🚀 

(Code review comments separately)

invenio_records_resources/resource_unit.py
```diff
@@ -6,12 +6,11 @@
 # modify it under the terms of the MIT License; see LICENSE file for more
 # details.
 
-"""State API."""
+"""Resource unit."""
```
ppanero, line 9: minor: since there are more than one in the file `resource_units.py`? (final *s*) It's true we have no rule on this (or at leas we don't follow). But files with more than one are e.g. `responses.py` files with one `service.py`. OCD kicked in sorry :)

invenio_records_resources/services/file_metadata_service.py
```diff
@@ -0,0 +1,63 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+# Copyright (C) 2020 Northwestern University.
+#
+# Invenio-Records-Resources is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Record Service API."""
```
ppanero, line 10: minor: file naming again. If we are in the `service` folder IMHO there is no need to put `service` in the name. See views, resources, etc. they all are `records` etc. no `records_view` or `records_resource`. (Just the file name to make them shorter which helps on imports).

invenio_records_resources/services/file_metadata_service.py
```diff
@@ -0,0 +1,63 @@
...
+
+from invenio_records_permissions.policies.records import RecordPermissionPolicy
+
+from ..resource_unit import IdentifiedRecord, RecordSearchState
+from .service import Service, ServiceConfig
+
+
+class FileMetadata:
+    """Fake File metadata resource unit."""
```
ppanero, line 19: minor: if its fake for now I wouldn't put it in the docstring cuz we are gonna forget to remove it. I'd say a todo or nothing.

invenio_records_resources/services/__init__.py
```diff
@@ -9,10 +9,17 @@
...
 from .data_validator import DataValidator, MarshmallowDataValidator
-from .service import RecordService, RecordServiceConfig
+from .file_metadata_service import FileMetadataService, \
+    FileMetadataServiceConfig
+from .file_service import FileService, FileServiceConfig
+from .record_service import RecordService, RecordServiceConfig
 
 __all__ = (
     "DataValidator",
+    "FileMetadataService",
```
ppanero, line 19: moderate: would be nice to export the generic service interface so the import on others (e.g. drafts) is not Nile-long hehe :)

invenio_records_resources/resource_unit.py
fenekku: :100: You are absolutely right. I did it for `service/` -> `services/` but forgot this one. On it!

invenio_records_resources/services/file_metadata_service.py
fenekku: Was trying to see if it made it easier to find the file in my text editor, but it doesn't make a difference really. Renaming!

invenio_records_resources/services/file_metadata_service.py
fenekku: Going for `TODO`

invenio_records_resources/services/__init__.py
fenekku: Ha! :) Included `Service` + `ServiceConfig`

PR #40, opened 2020-07-20
errors: handle pid related resolver errors

zzacharo:


invenio_records_resources/resources/record.py
```diff
@@ -53,6 +55,21 @@ class RecordResourceConfig(ResourceConfig):
...
+            )
+        ),
+        PIDUnregistered: create_errormap_handler(
+            HTTPJSONException(
+                code=404,
+                description="The pid is not registered.",
+            )
+        ),
+        PIDRedirectedError: create_pidredirected_error_handler(
+            route="/records/{pid_value}"
```
fenekku, line 71: normal: can this be replaced with `RecordResourceConfig.item_route` ?

tests/test_record_resource.py
```diff
@@ -21,6 +21,39 @@
 HEADERS = {"content-type": "application/json", "accept": "application/json"}
 
 
+@pytest.fixture(scope="function")
+def app_with_custom_minter(app):
+    """Application providing a minter creating unregistered pid values."""
+    from invenio_pidstore.minters import recid_minter
+    from invenio_pidstore.models import PIDStatus, RecordIdentifier
+    from invenio_pidstore.providers.recordid import RecordIdProvider
+    from invenio_pidstore.proxies import current_pidstore
```
fenekku, line 30: minor: These can be placed at the top-level. It's fine.

tests/test_record_resource.py
```diff
@@ -150,3 +183,61 @@ def test_read_deleted_record(client, input_record):
...
+    response = client.get("/records/{}".format(recid), headers=HEADERS)
+    assert response.status_code == 404
+
+    assert response.json["status"] == 404
+    assert response.json['message'] == "The pid is not registered."
+
+
+def test_read_record_with_redirected_pid(client, input_record):
+    """Test read a record with a redirected pid."""
+    from invenio_pidstore.models import PersistentIdentifier
```
fenekku, line 219: Move to top-level

Nested imports are scary :scream_cat: :
- either they exist *because* there are circular dependencies in the codebase which is really not good
- or they *cause* circular dependencies because developers are then not aware of the circular imports

It's a small thing that helps so much in the long run :)
(there are also performance implications, but performance is way less important than the above)

invenio_records_resources/resources/record.py
zzacharo: It is not needed so I removed it :)

invenio_records_resources/errors.py
```diff
@@ -10,10 +10,40 @@
 
 import json
 
+from flask import abort, current_app, jsonify, make_response, request, url_for
```
ppanero, line 13: Remove unused imports

```suggestion
from flask import jsonify, make_response, request, url_for
```

invenio_records_resources/errors.py
```diff
@@ -10,10 +10,40 @@
...
 from invenio_rest.errors import RESTException, RESTValidationError
+from werkzeug.routing import BuildError
 
 # TODO: Revise this
```
ppanero, line 17: Remove old comment:

```suggestion
```

invenio_records_resources/errors.py
```diff
@@ -10,10 +10,40 @@
...
 
+
+def create_pidredirected_error_handler(route):
```
ppanero, line 20: moderate:  below is called `pid_redirected`, consistency wise I would say both with _ or without (my take is with)

```suggestion
def create_pid_redirected_error_handler(route):
```

invenio_records_resources/errors.py
```diff
@@ -10,10 +10,40 @@
...
+    """Creates an error handler for `PIDRedirectedError` error."""
+
+    def pid_redirected_error_handler(e):
+        try:
+            # Check that the source pid and the destination pid are of the same
+            # pid_type
+            assert e.pid.pid_type == e.destination_pid.pid_type
+            # Redirection works only for the item route of the format
+            # `/records/<pid_value>`
+            location = url_for(
```
ppanero, line 30: shelvable: at business layer we are trying to build links differently (to avoid ties to views), maybe we should open an issue to have this consistent accross the codebase.

zzacharo: I went for the request as the view name is not retrievable from the `ResourceConfig`. Still, using the `request` is fine as we are handling errors that we want to return a response back. Which is your take on how to ideally build this?

ppanero: I dont have one, that's why `shelve`. My point was more on "this is yet another case of inconsistency" should we list them in an issue concerning "link building"?

zzacharo: Ok I will create a general one regarding link building :)

zzacharo: @ppanero https://github.com/inveniosoftware/invenio-records-resources/issues/45. You can add more code cases there :)

PR #43, opened 2020-07-21
Service refactor: syntactic sugar

### LINKED ISSUE #36: services: syntactic sugar
Syntactic sugar: isn't it nicer to have minter, data_validator as properties in the RecordService class so you can write e.g self.data_validator.validate(data)?

ppanero:
closes #36 

- Fix create test error code. Now 201.
- Make function attributes into singleton properties (makes code smaller and cleaner :) )

### DIFF
--- a/invenio_records_resources/services/record.py
+++ b/invenio_records_resources/services/record.py
@@ -117,14 +117,12 @@ class RecordService(Service):
         if not lt_es7:
             extras["track_total_hits"] = True
 
-        self.search_engine.search_arguments(
-            pagination=pagination,
-            extras=extras
-        )
-
         # Parse query and execute search
         query = self.search_engine.parse_query(querystring)
-        search_result = self.search_engine.execute_search(query)
+        search_result = self.search_engine.search_arguments(
+            pagination=pagination,
+            extras=extras
+        ).execute_search(query)
 
         # Mutate search results into a list of record states
         record_list = []
--- a/invenio_records_resources/services/record.py
+++ b/invenio_records_resources/services/record.py
@@ -54,10 +54,7 @@ class RecordService(Service):
     #
     # Low-level API
     #
-
-    #
-    # Persistent identifier resolution
-    #
+    @property
     def resolver(self):
         """Factory for creating a resolver instance."""
         return self.config.resolver_cls(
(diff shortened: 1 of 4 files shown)


invenio_records_resources/services/record.py
```diff
@@ -49,52 +49,73 @@ class RecordService(Service):
...
+    _search_engine = None
 
     #
     # Low-level API
     #
-
-    #
-    # Persistent identifier resolution
-    #
+    @property
```
zzacharo, line 62: nit: I think is easier to use `functools.cached_property` :) You will avoid all the `_` attributes for singleton...

ppanero: Nice! 🤝

invenio_records_resources/services/record.py
```diff
@@ -53,48 +54,48 @@ class RecordService(Service):
     #
     # Low-level API
     #
-
-    #
-    # Persistent identifier resolution
-    #
+    @cached_property
```
fenekku, line 57: MAJOR: Hey, I know this was well-intentioned, but... we should *not* do this. :)

There are 2 hard things in computer science: naming things and cache invalidation. This introduces a cache, that needs to be explicitly invalidated. In particular it prevents changing the config dynamically without having to call https://werkzeug.palletsprojects.com/en/1.0.x/utils/#werkzeug.utils.invalidate_cached_property . From hard-experience, developers down the road will not realize this and become extremely frustrated ;) 


Try this:

```python
from werkzeug.utils import cached_property


class ConfigA:
    a = 'A'


class ConfigB:
    b = 'B'


class A:

    def __init__(self, config):
        self.config = config

    @cached_property
    def foo(self):
        print("fooing...")
        return self.config.a

a1 = A(ConfigA)
print("a1.foo()", a1.foo)
print("a1.foo()", a1.foo)

print("changing config")
a1.config = ConfigB

print("a1.foo()", a1.foo)
print("a1.foo()", a1.foo)
```

ppanero: My personal take on this is that "we should not change dynamically the config". If we want to change it dynamically we should rather use two different objects...

IMHO that's one of the advantages of OOP, otherwise we can simply code functionally and we change things on the spot...

```
from werkzeug.utils import cached_property

class ConfigA:
    a = 'A'


class ConfigB:
    a = 'B'

class A:
    def __init__(self, config):
        self.config = config

    @cached_property
    def foo(self):
        return self.config.a


a1 = A(ConfigA)
print("a1.foo", a1.foo)

a2 = A(ConfigB)

print("a2.foo", a2.foo)
```

Resulting in:
```
$ python test.py
a1.foo A
a2.foo B
```

lnielsen: Let's discuss IRL during stand-up today.

fenekku: Oh, I am sorry about the request change, I did not mean it as an injunction :(. This was a case of PTSD for me and I needed to press on an abort button :sweat_smile: : I wanted to make sure this was caught before merging since there was a previous approval. The smileys were not enough to convey my good faith. If we should not use that button at all, I will refrain from that. I apologize for the reaction.

Let's talk about the issue after standup and how I can convey good faith.

ppanero:
@fenekku as discussed IRL changes made. Could you approve? :)

PR #44, opened 2020-07-21
service: set default pid_type to recid_v2

ppanero:


ppanero:
This is the wrong approach. #46 is the way to go.

PR #46, opened 2020-07-21
split pid types

ppanero:
requires #36

There is a need to encapsulate pids management (resolver, minter, fetcher, types, even versioning). It will be addressed in a different PR.


ppanero:
requires #36

PR #48, opened 2020-07-23
SHOW ONLY: DO NOT MERGE

ppanero:
**ONLY `test_create_read_record` IS FULLY FUNCTIONAL**

**Idea:**

- An enriched `Record` class that returns always a `unit`. In this case PID + Record. 
- The pid manager will encapsulate anything that has to do with: resolver, fetcher, minters and providers.

**Problems:**
- Some attributes like `default_status` and `default_status_with_obj` are class attributes on `Providers`, which does not make it easy to set on demand.
- Many of the operations (e.g. pid creation) could be useful to do separately. However accesing Provider + Minter, defeats the purpose of the Minter. --> Up to what extent do we want to not change PIDStore?
- Data validator module and classes si part of the `service` atm. However, it is imported in the `pid_record`. Solution: move to the `record` module, which would make sense. Do the same with `schemas`.

**Alternative:**
- An instantiable class with `pid` and `record` attributes.
- Problems most likely would persist on this approach.

### DIFF
--- a/invenio_records_resources/record/__init__.py
+++ b/invenio_records_resources/record/__init__.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2020 CERN.
+#
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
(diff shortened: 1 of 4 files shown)


lnielsen:
Supersede by ``sep``-branch and system fields.

PR #50, opened 2020-07-24
fix: test search correctly

### LINKED ISSUE #19: tests: fix search tests
**Current**
Tests are not being cleaned correctly: indexes are not created/deleted properly. A fake index must be setup and `base_app` must be used rather than `app` in the fixtures. 

**Previous**
The get (search) endpoint is failing because the fetcher can't retrieve `'recid'` key from returned `data`:

```
invenio_resources/service/service.py:158: in search
    pid = cls.fetcher()(data=record, record_uuid=None)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

record_uuid = None
data = {'_access': {'metadata_restricted': False, 'files_restricted': False}, '_owners': [1], '_created_by': 1, 'access_right...d title', 'type': 'Other', 'lang': 'eng'}], 'publication_date': '2020-07-08', '_publication_date_search': '2020-07-08'}

    def recid_fetcher(record_uuid, data):
        """Legacy way to fetch a record's identifiers.
    
        :param record_uuid: The record UUID.
        :param data: The record metadata.
        :returns: A :data:`invenio_pidstore.fetchers.FetchedPID` instance.
        """
        pid_field = current_app.config['PIDSTORE_RECID_FIELD']
        return FetchedPID(
            provider=RecordIdProvider,
            pid_type=RecordIdProvider.pid_type,
>           pid_value=str(data[pid_field]),
        )
E       KeyError: 'recid'

.venv/lib/python3.6/site-packages/invenio_pidstore/fetchers.py:67: KeyError
```

Needs investigation.

fenekku:
close #19 

`test_read_record_with_different_type_of_redirected_pid` is now failing on my machine... It actually raises the exception instead of serializing to a 500. I know pidstore and error handling have been modified recently. What am I missing?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,12 +30,14 @@ services:
 
 env:
   global:
+    - ES6_DOWNLOAD_URL="https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-6.8.2.tar.gz"
     - ES7_DOWNLOAD_URL="https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-7.3.0-linux-x86_64.tar.gz"
     - ES_HOST=127.0.0.1
     - SQLALCHEMY_DATABASE_URI="postgresql+psycopg2://postgres@localhost:5432/invenio"
   matrix:
     # - REQUIREMENTS=lowest ES_URL=$ES7_DOWNLOAD_URL
     # - REQUIREMENTS=release ES_URL=$ES7_DOWNLOAD_URL
+    - REQUIREMENTS=devel ES_URL=$ES6_DOWNLOAD_URL EXTRAS=all,postgresql,elasticsearch6
     - REQUIREMENTS=devel ES_URL=$ES7_DOWNLOAD_URL EXTRAS=all,postgresql,elasticsearch7
 
 python:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2020 CERN.
+# Copyright (C) 2020 Northwestern University.
 #
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -38,7 +39,7 @@ env:
     # - REQUIREMENTS=lowest ES_URL=$ES7_DOWNLOAD_URL
     # - REQUIREMENTS=release ES_URL=$ES7_DOWNLOAD_URL
     - REQUIREMENTS=release ES_URL=$ES6_DOWNLOAD_URL EXTRAS=all,postgresql,elasticsearch6
-    - REQUIREMENTS=release ES_URL=$ES7_DOWNLOAD_URL EXTRAS=all,postgresql,elasticsearch7 DEPLOY=true
+    - REQUIREMENTS=release ES_URL=$ES7_DOWNLOAD_URL EXTRAS=all,postgresql,elasticsearch7
     - REQUIREMENTS=devel ES_URL=$ES7_DOWNLOAD_URL EXTRAS=all,postgresql,elasticsearch7
 
 python:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ matrix:
   fast_finish: true
   allow_failures:
     # To allow failures, you need to specify the full environment
-    - env: REQUIREMENTS=devel
+    - env: REQUIREMENTS=devel ES_URL=$ES7_DOWNLOAD_URL EXTRAS=all,postgresql,elasticsearch7
 
 cache:
   - pip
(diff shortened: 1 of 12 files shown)


tests/mock_module/__init__.py
```diff
@@ -0,0 +1,9 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2017-2018 CERN.
```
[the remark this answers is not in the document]
fenekku: Yes! (this was taken from invenio-records-rest :sweat_smile: )

tests/mock_module/mappings/v6/__init__.py
```diff
@@ -0,0 +1,9 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2017-2018 CERN.
+#
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Elasticsearch version 6 mappings."""
```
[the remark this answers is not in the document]
fenekku: Oh, interesting. Since we have version dependent code, I am adding ES6 to our testing matrix.

fenekku:
Can be re-reviewed now.

.travis.yml
```diff
@@ -19,7 +20,7 @@ matrix:
   fast_finish: true
   allow_failures:
     # To allow failures, you need to specify the full environment
-    - env: REQUIREMENTS=devel
+    - env: REQUIREMENTS=devel ES_URL=$ES7_DOWNLOAD_URL EXTRAS=all,postgresql,elasticsearch7
```
ppanero, line 23: question: It is not enough with `REQUIREMENTS=devel`  it requires the full matrix value to "identify" it?

.travis.yml
```diff
@@ -56,7 +60,8 @@ before_install:
 
 install:
   - "travis_retry pip install -r .travis-${REQUIREMENTS}-requirements.txt"
-  - "travis_retry pip install -e .[all]"
+  - "travis_retry pip install -e .[$EXTRAS]"
+  - "pip freeze"
```
ppanero, line 64: 🚀

setup.py
```diff
@@ -45,9 +46,12 @@
     "tests": tests_require,
 }
 
-extras_require["all"] = []
-for reqs in extras_require.values():
-    extras_require["all"].extend(reqs)
+all_requires = []
```
ppanero, line 49: doubt: why do we need to skip elasticsearch extras? Is this the reason some were "forcefully" being installed (always 6 and then move to 7, if 7 was specified?)

tests/conftest.py
```diff
@@ -46,7 +57,8 @@ def app_config(app_config):
     https://github.com/inveniosoftware/invenio-records-permissions/issues/51
     """
     app_config["RECORDS_REST_ENDPOINTS"] = {}
-
+    app_config["INDEXER_DEFAULT_DOC_TYPE"] = "testrecord"
```
ppanero, line 60: normal/enhancement: we can set up also [`SEARCH_MAPPINGS`](https://invenio-search.readthedocs.io/en/latest/configuration.html#invenio_search.config.SEARCH_MAPPINGS) variable. That way, only this mapping will be created and not also the others specified in other modules entrypoints (Im quite sure the records-... is being created anyway).

.travis.yml
fenekku: Yes, I noticed that none of the `env` showed in the "Allowed to fail" section and then I noticed the comment above saying that the full environment had to be passed. Once the full environment was passed, that one was allowed to fail :smile:

.travis.yml
fenekku: That `pip freeze` pattern you introduced is really helpful. Thanks! Just making sure we have it here too.

tests/conftest.py
fenekku: I was thinking once we sever ties with invenio-records-rest/invenio-files-rest, we won't have other modules creating mappings that this module depends on. This should resolve itself with https://github.com/inveniosoftware/invenio-records-permissions/issues/51 and once we do files next month.

setup.py
fenekku: This took me a while to discover... When we run `pip install -e .[all, elasticsearch7]` as before, because `all` extra lists `elasticsearch6` as a dependency, it would get installed and override the installation of elasticsearch7 (or vice-versa - no install guarantees because both are dependencies). All other packages (e.g. https://github.com/inveniosoftware/invenio-records-rest/blob/master/setup.py#L68) that use the elasticsearch extras, go through this filtering process to prevent this from happening.

Issue #55, opened 2020-07-29
Pass UI routes to Service


fenekku:
The service layer needs to know about the UI routes in order to provide links to HTML page of records.

Right now it assumes the UI route is the API route without the `api/` part which is obviously not always the case.

Related to #45

fenekku:
closed via #62

Issue #39, opened 2020-07-20
tombstone: add record serialization in error response


zzacharo:
When a [PidDeletedError](https://github.com/inveniosoftware/invenio-pidstore/blob/master/invenio_pidstore/errors.py#L45) is raised, a `record` is associated with the error. We should be able to serialize somehow this record in the error response so users can return any information they want. We need to investigate how to serialize this response.

PR #62, opened 2020-08-11
links: use linker

fenekku:
First PR in the link generation trio to be ready.


tests/services/test_links.py
```diff
@@ -0,0 +1,65 @@
...
+@pytest.fixture()
+def identity_no_need():
+    """Simple identity fixture."""
+    i = Identity(1)
+    # i.provides.add(UserNeed(1))
+    # i.provides.add(Need(method='system_role', value='any_user'))
+    return i
+
+
+def test_record_links(app, identity_simple, input_service_data, es):
```
ppanero, line 36: Moderate: Missing docstring? I noticed in my local things do not fail with `./run-tests` it does not pick this things, needs investigation (separatedly, shelved).

tests/services/test_links.py
```diff
@@ -0,0 +1,65 @@
...
+
+    expected_links = {
+        "self": f"https://localhost:5000/api/records/{pid_value}",
+        "self_html": f"https://localhost:5000/records/{pid_value}",
+        "delete": f"https://localhost:5000/api/records/{pid_value}",
+    }
+    assert expected_links == links
+
+
+def test_permission_record_links(
```
ppanero, line 54: Moderate: Missing docstring?

tests/services/test_links.py
```diff
@@ -0,0 +1,65 @@
...
+    # i.provides.add(Need(method='system_role', value='any_user'))
+    return i
+
+
+def test_record_links(app, identity_simple, input_service_data, es):
+    record_service = RecordService()
+    record_unit = record_service.create(input_service_data, identity_simple)
+    pid_value = record_unit.id
+
+    links = record_service.linker.links(
```
ppanero, line 41: knowledge check: Since the links comes back from *creation* (in the unit). Why is the linker called again? My point being the test is:
- Creating a record (that comes with links)
- Creating the links again (in the same fashion)
- Comparing the second part to the expected.

I understand this is testing the `linker`, so the assertion is correct. However, for completeness we could check also the links are the same that came with the record? (though that might be a test in the service?)

``` python
assert expected_links == links == record_unit.links
```

It's okay, is just the test is not fully clear to me.

invenio_records_resources/links.py
```diff
@@ -135,3 +119,113 @@ def links(self):
...
+
+        :param link_builders: dict(string, list<LinkBuilder>)
+        """
+        self.link_builders = link_builders
+
+    def links(self, namespace, identity, **kwargs):
+        """Returns dict of links."""
+        output_links = {}
+        for link_builder in self.link_builders[namespace]:
+            link = link_builder.route_to_link(identity, **kwargs)
```
ppanero, line 141: minor: This is calling a function that it does not implement. Should it have a default behaviour? or is ok to let it fail? If the second, maybe is a good idea to put in the docstring something like "It also requires a link builder..."

invenio_records_resources/links.py
```diff
@@ -135,3 +119,113 @@ def links(self):
...
+        else:
+            return {}
+
+
+class RecordSelfLinkBuilder(RecordLinkBuilder):
+    """Builds record "self" link."""
+
+    def __init__(self, config):
+        """Constructor."""
+        super(RecordSelfLinkBuilder, self).__init__(
```
ppanero, line 175: normal: There are four classes just overwriting the constructor. Could they just be instances?

```python
record_self_link_builder = RecordLinkBuilder(key="self",
             route=api_route(config.record_route),
             action="read",
             permission_policy=config.permission_policy_cls
         )
```

Same for the few below. I think there is no need to create classes here. In the end if someone needs a change, they will create another instance.

invenio_records_resources/services/record.py
```diff
@@ -97,16 +121,6 @@ def record_cls(self):
         """Factory for creating a record class."""
         return self.config.record_cls
 
-    @property
```
ppanero, line 100: :fire: yes!!!!

invenio_records_resources/services/record.py
```diff
@@ -118,8 +132,10 @@ def read(self, id_, identity):
         """Retrieve a record."""
         pid, record = self.resolve(id_)
         self.require_permission(identity, "read", record=record)
-        links = ResourceUnitLinks(self.item_route, pid.pid_value).links()
-        # Todo: how do we deal with tombstone pages
+        links = self.linker.links(
+            "record", identity, pid_value=pid.pid_value, record=record
+        )
+        # TODO: how do we deal with tombstone pages
```
ppanero, line 138: Hasn't this been dealt with by @zzacharo (old comment that should be removed?)

tests/conftest.py
```diff
@@ -98,6 +98,12 @@ def app_config(app_config):
     app_config["RECORDS_REST_ENDPOINTS"] = {}
     app_config["INDEXER_DEFAULT_DOC_TYPE"] = "testrecord"
     app_config["INDEXER_DEFAULT_INDEX"] = TestSearch.Meta.index
+    app_config["RECORDS_UI_ENDPOINTS"] = {
```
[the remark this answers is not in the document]
fenekku: Maybe? :) Taking the UI route from there does indeed close: #55 . 

It would be nicer if routes were defined together, but I don't think the Resource should know the UI route right now... 

I had a shower-thought about actually using the Service as the source of truth for routes. It would require some little changes, but that's a possibility to simplify where routes should be and prevent potential duplication...

tests/services/test_links.py
fenekku: Docstrings are not needed for tests (by our linters). Personally I only exceptionally put a docstring in tests if there are more details than can be conveyed by the test name.

tests/services/test_links.py
fenekku: same

tests/services/test_links.py
fenekku: Ah! I was also wondering at it this morning... and then I remembered. I will add the below to make it clearer what I was going for.

```
    # NOTE: We are testing linker.links() as opposed to record_unit.links
    #       because the former is used in all RecordService methods (not just
    #       create)
```

Essentially, it's to make the test unbiased to the `create()` method in particular, by testing the underlying method that is used in all RecordService methods. Then invenio-rdm-records' tests check all the ramifications.

invenio_records_resources/services/record.py
fenekku: Oh git is weird: I just capitalized the existing `Todo` to `TODO` and it makes it show up like a new line... I haven't touched this aspect otherwise, so I would leave it until we finalize tombstones (#9 / #39)

invenio_records_resources/links.py
fenekku: The `config` is dynamically provided at application runtime, so we can't assign it beforehand. Writing this in-place is a bit too much as well. (and finally I am not a fan of global assignments if I can do it another way :cat:  )

invenio_records_resources/links.py
fenekku: Not sure I understand this one... Each link_builder is expected to have a `route_to_link` method. If it doesn't this should just fail.

PR #63, opened 2020-08-13
data validation: allow soft (partial) validation

ppanero:


tests/services/test_data_validator.py
```diff
@@ -70,6 +71,18 @@ def test_marshmallow_data_validator_with_custom_schema():
     assert "custom_field" in validated_data
 
 
+def test_marshmallow_data_validator_with_custom_schema_partial():
+    """Tests the marshmallow data validator."""
+
+    input_record = {"extra_field": "test_value"}
+
+    validator = MarshmallowDataValidator(schema=CustomSchemaJSONV1)
+    validated_data = validator.validate(data=input_record, partial=True)
```
zzacharo, line 80: maybe just add a case before, showing what happens with partial=False i.e requiring the `custom_field`? WDYT? Just to have a clear view of what is happening in both cases...

ppanero: Thats the test below no?

ppanero: Done :)

Issue #67, opened 2020-08-19
Allow for per instance configuration of ResourceConfig / ServiceConfig


fenekku:
Our approach so far makes it very easy for a module like invenio-rdm-records to define its own BibliographicResourceConfig or BibliographicServiceConfig that (eventually) inherit from the config defined here in records-resources. However this approach doesn't take into account the use case where an instance might want to further customize these configuration objects.

For instance, changing the API urls is not possible per instance. 

This is tricky to accomplish because `.as_blueprint()` is called at the top-level in `views.py` *outside of an application context*...

Potential solutions:
- make sure nothing in configs and services' constructors is relying on the current_app (might sometimes be tricky)
- place blueprint registration in ext.py of invenio-rdm-records (like invenio-records-rest). To avoid automatic endpoint registration, we may want to have a whitelist only policy that is enabled via entry_points... (doesn't feel like we are improving much on records-rest here, but maybe that's the way to go with our constraints) 

**EDIT**
- I think providing a callable blueprint is actually the way to go, this way a `ResourceConfig` can be defined in the config.py and fetched with an application context. (I think).

PR #76, opened 2020-09-02
config: make customizable via config.py

fenekku:
- part of #67
 - change in rdm-records needed to fully close the issue


fenekku:
Merging to get things moving.

PR #83, opened 2020-09-09
facet: port aggregations and post_filters from records-rest

fenekku:


fenekku:
requires release of flask-resources to pass

PR #84, opened 2020-09-10
records: new data access layer and a PIDField system field.

lnielsen:
- Depends on:
  - inveniosoftware/invenio-records#225
  - inveniosoftware/invenio-indexer#122 
  - inveniosoftware/pytest-invenio#58

- Refactor tests directory to split into presentation, service and data layers.

- Adds PIDField system field as a replacement for minters.

- Adds a data access layer based on the new Invenio-Records.


tests/records/test_api.py
```diff
@@ -0,0 +1,104 @@
...
+    # Empty record creation works, and injects a schema.
+    record = Record.create({})
+    db.session.commit()
+    assert record.schema
+
+    # JSONSchema validation works.
+    pytest.raises(ValidationError, Record.create, {'metadata': {'title': 1}})
+
+
+def test_record_via_field(app, db):
```
slint, line 58: Should the PIDField also be tested here?

slint (commented):
LGTM, some minor comments on the PIDField config

PR #86, opened 2020-09-11
Refactoring

lnielsen:


invenio_records_resources/resources/record.py
```diff
@@ -49,55 +67,49 @@ def search(self):
         }
         pagination = request_args
 
+        # TODO: Args parsing has to allow for easily adding new parameters,
+        # when someone extends the search engine.
```
fenekku, line 71: - SearchURLArgsSchema() does this for URL search args, I think... i.e perhaps passing `request_args` and letting search_engine deal with it?

fenekku (commented):
Comments so far. I have stopped at record_config.py (non-inclusive). Will continue tomorrow.

fenekku (commented):
We talked about most points. Done reviewing. Thanks for all this!

Issue #37, opened 2020-07-17
tests: remove app config when records-permissions is not dependent on records-rest


ppanero:
https://github.com/inveniosoftware/invenio-records-resources/blob/efc5af213e79057ecf66135c7ac27a54b841af40/tests/conftest.py#L35

Depends on https://github.com/inveniosoftware/invenio-records-permissions/issues/51

PR #88, opened 2020-09-15
migrate to new ci features

ppanero:
Changes:

- Migrate to GH action releases
- Migrate to docker-services-cli

Closes https://github.com/inveniosoftware/invenio-records-resources/issues/37

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -1,0 +1,26 @@
+on:
+  push:
+    tags:
+      - v*
+
+jobs:
+  build-n-publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install setuptools wheel
+      - name: Build package
+        run: |
+          python setup.py compile_catalog sdist bdist_wheel
+      - name: pypi-publish
+        uses: pypa/gh-action-pypi-publish@v1.3.1
+        with:
+          user: __token__
+          password: ${{ secrets.pypi_token }}
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,6 @@ branches:
   except:
   - /^v\d+\.\d+(\.\d+)?(\S*)?$/
 
-addons:
-  postgresql: 9.5
-
 notifications:
   email: false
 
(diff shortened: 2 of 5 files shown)


.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,26 @@
...
+          python -m pip install --upgrade pip
+          pip install setuptools wheel
+      - name: Build package
+        run: |
+          python setup.py compile_catalog sdist bdist_wheel
+      - name: pypi-publish
+        uses: pypa/gh-action-pypi-publish@v1.3.1
+        with:
+          user: __token__
+          password: ${{ secrets.pypi_token }}
```
fenekku, line 26: Just after you left for vacation there was an issue where additional manual steps were required on GitHub to make this work (something to do with the token and its scope I think). Did you do those already?

run-tests.sh
```diff
@@ -6,8 +6,12 @@
...
-pydocstyle invenio_records_resources tests docs && \
-isort invenio_records_resources tests --check-only --diff && \
-check-manifest --ignore ".travis-*" && \
-sphinx-build -qnNW docs docs/_build/html && \
-pytest
+python -m pydocstyle invenio_records_resources tests docs && \
+python -m isort invenio_records_resources tests --check-only --diff && \
+python -m check_manifest --ignore ".travis-*" && \
+python -m sphinx.cmd.build -qnNW docs docs/_build/html && \
+docker-services-cli up es postgresql redis
```
fenekku, line 13: I've only needed redis when testing invenio-app-rdm (other modules didn't need it).

postgresql I would only run on travis (local tests use sqlite). 

Maybe they can be passed as arguments to not have to spin them up?

fenekku (approved):

.travis.yml
```diff
@@ -20,27 +21,18 @@ matrix:
...
-    - ES_HOST=127.0.0.1
-    - SQLALCHEMY_DATABASE_URI="postgresql+psycopg2://postgres@localhost:5432/invenio"
+    - DOCKER_COMPOSE_VERSION=1.26.2
   matrix:
-    # - REQUIREMENTS=lowest ES_URL=$ES7_DOWNLOAD_URL
-    # - REQUIREMENTS=release ES_URL=$ES7_DOWNLOAD_URL
-    - REQUIREMENTS=release ES_URL=$ES6_DOWNLOAD_URL EXTRAS=all,postgresql,elasticsearch6
-    - REQUIREMENTS=release ES_URL=$ES7_DOWNLOAD_URL EXTRAS=all,postgresql,elasticsearch7
-    - REQUIREMENTS=devel ES_URL=$ES7_DOWNLOAD_URL EXTRAS=all,postgresql,elasticsearch7
+    - REQUIREMENTS=release EXTRAS=all,postgresql,elasticsearch6 ES_VERSION=ES_6_LATEST
```
lnielsen, line 33: Comment: We should add MySQL as well.

run-tests.sh
lnielsen: Related to https://github.com/inveniosoftware/docker-services-cli/issues/7

lnielsen:
I assume this would be easy to port to the ``sep`` branch?

.github/workflows/pypi-publish.yml
ppanero: I'm not aware of said troubles 😅 *before* I left there was the need of setting it up, but it was set up and it worked for `docker-services-cli` already. Is this what you meant?

run-tests.sh
ppanero: > I've only needed redis when testing invenio-app-rdm (other modules didn't need it).
Hm... could be, I'll try. It was set on the travis file and I sort of assume it was needed.

> postgresql I would only run on travis (local tests use sqlite).
As Lars mentioned it is a known issue. I'd shelve it for when we have time to go back to `docker-services-cli`.

.travis.yml
ppanero: Since there is not `model` defined here we did not find it necessary. `Invenio-RDM-Records` should do the checking of the different databases?

.github/workflows/pypi-publish.yml
slint: It was a naming issue, since in the workflow the token was named `pypi_password`, while in the org settings it was `pypi_token`... Now both exist, but we should align to use one (I prefer `pypi_token`)

PR #99, opened 2020-09-18
Sync sep master

zzacharo:


invenio_records_resources/services/records/params/querystr.py
```diff
@@ -28,18 +28,19 @@ def parse(self, query_str):
 class QueryStrParam(ParamInterpreter):
     """Evaluate the 'q' parameter."""
 
-    def __init__(self, parser_cls=QueryParser):
-        """Initialize with the query parser."""
-        self.parser_cls = parser_cls
+    # def __init__(self):
+    #     """Initialize with the query parser."""
+    #     self.parser_cls = config.search_query_parser_cls
```
fenekku, line 33: Comments can be removed

invenio_records_resources/services/records/params/sort.py
```diff
@@ -8,24 +8,48 @@
...
+                            self.config.search_sort_default_no_query)
+                    )
             return search
 
-        # TODO
+        if sort not in sort_options:
+            raise ValidationError("Invalid sort parameter.")
+
+        search = search.sort(*self.get_option_fields(sort))
+        params['sort'] = sort
```
fenekku, line 53: Can remove, already in there (or not)

invenio_records_resources/services/records/params/facets.py
```diff
@@ -16,11 +16,27 @@
 class FacetsParam(ParamInterpreter):
```
fenekku, line 16: Name emphasis should be on Interpreter: `FacetsInterpreter` rather than `Params` (these classes interpret the params to give a modified search object). This is in keeping with how they are used: https://github.com/inveniosoftware/invenio-records-resources/pull/99/files#diff-6287c413b00ac98631d623e0329c9f0dR97

tests/services/test_service.py
```diff
@@ -68,3 +69,38 @@ def test_simple_flow(app, service, identity_simple, input_data):
...
+    assert list(res.hits)[0] == first_item.data
+
+    # Search with existing sort parameter
+    res = service.search(identity_simple, sort="mostrecent")
+
+    assert res.total == 2
+    assert list(res.hits)[0] == second_item.data
+
+    # Search with non existing sort parameter
+    pytest.raises(ValidationError, service.search, identity_simple, sort="foo")
```
fenekku, line 106: Perhaps I am biased, but would simply copy *as-is* : https://github.com/inveniosoftware/invenio-records-resources/blob/master/tests/test_sorting.py 

I specifically made these tests at the endpoint level, because I was expecting the internals to change, but the ouput to stay the same from Lars+Alex's work. If I may say so myself (:wink: ), this is the perfect way to make sure we kept the API guarantees we provide through change to the code base (without having to rewrite tests either).

invenio_records_resources/services/records/params/sort.py
```diff
@@ -8,24 +8,48 @@
...
+        sort_options = self.config.search_sort_options
+        return sort_options[option].get("fields", [])
 
     def apply(self, identity, search, params):
         """Evaluate the query str on the search."""
+        sort_options = self.config.search_sort_options
         sort = params.get('sort')
         if sort is None:
+            if self.has_q(params):
+                if self.config.search_sort_default:
```
fenekku, line 36: Hmmm very nested ;) I can look into that.

fenekku (commented):
I will deal with the comments I made and add faceting / pagination back

invenio_records_resources/services/records/params/facets.py
lnielsen: For me the emphasis is on the param. That's why I named the package ``params`` and went with the suffix ``Param``.

tests/services/test_service.py
lnielsen: The service is a high-level programmatic API that shouldn't be changing and breaking changes needs to be caught. I also prefer that each layer is well-tested on the features they provide - in this case it's the service that provides the sorting feature, while the resource "just" provide parameter parsing as this makes catching bugs easier - if the test is too high-level you don't know where in the stack you have too look. 
Invenio-RDM-Records should add the high-level tests that goes through the entire application stack and ensures that RDM stays with a stable API.

fenekku: I'll concede the addition of tests at the service level to ensure the programmatic interface of the service. But I would keep the endpoint level tests (enabled a number of them already), because records-resources does provide the core resource + service functionality. SImilar tests in rdm-records are fine, but if they break the fix is not in records-resources; whereas, tests here are great because if they break, the fix is in the same repository.

PR #110, opened 2020-09-22
109 fix links generation

fenekku:


tests/mock_module/resource.py
```diff
@@ -12,14 +12,16 @@ class ResourceConfig(RecordResourceConfig):
     list_route = "/mocks"
     item_route = f"{list_route}/<pid_value>"
 
+    # IDEA: Formalize the concept of thunk in our codebase and use it
+    # TODO: Changing list_route / item_route shouldn't require redefining these
```
lnielsen, line 16: I disagree with this point. I'd like to have links generation and app routing completely separated. This makes them decoupled components which are far easier to maintain.

invenio_records_resources/services/records/results.py
```diff
@@ -152,6 +152,7 @@ def pagination(self):
     @property
     def links(self):
         """Get the search result links."""
+        # TODO: Refactor because of magic below
```
lnielsen, line 155: We hold a reference to the same dictionary containing the links in 1) the link store and 2) the record projection. 

This allows us to replace links anywhere in the record projection no matter where they are. The LinksField and LinksSchema do the bookkeeping job.

If you were to replace this mechanism, you need another way of knowing all the places in the record projection where you have to go and update links. This tracking can very quickly become cumbersome and error-prone, whereas the solution of holding a reference is cleanly implemented and reliable. Thus, I'd rather call this advanced Python rather than magic.

invenio_records_resources/services/records/links.py
```diff
@@ -25,7 +25,7 @@ def pid_value_dict(record, context):
     return {'pid_value': record.pid.pid_value}
 
 
-def search_args_dict(page_offset):
+def search_link_params(page_offset):
```
lnielsen, line 28: 👍 Name is much better

tests/mock_module/resource.py
fenekku: Not sure I understand this one :) . App routing and link generation are already coupled: routing uses the `list_route`. And even if they were not "code" coupled, they are semantically coupled (not a bad coupling): we should be generating links to the endpoints defined via `iist_route` / `item_route`...

invenio_records_resources/services/records/results.py
fenekku: > I'd rather call this advanced Python rather than magic.
Ha! 

The search links are not currently using `LinksField`. Record projection is `RecordList` or `RecordItem` right? But even then:

The way I see it is to put this `LinksStore.resolve` call as part of the `LinksSchema.dump` flow, so that what we get actually has the links. My biggest worry is that a call to `dump` doesn't generate links but parameters i.e. Linkstore isn't storing links but parameters until the resolve call is made at which point the same variable that was holding parameters is now storing links... It can get more confusing if `LinksStore.add` is then called: we are now mixing the two... The "no matter where they are" part would also be solved because the schema knows.      

In `RecordItem` case, it could look like:

```python
    @property
    def data(self):
        """Property to get the record."""
        if self._data:
            return self._data

        links = LinksStore(host=_current_host, config=self._links_config)

        self._data = self._service.schema.dump(
            self._identity,
            self._record,
            pid=self._record.pid,
            record=self._record,
            links_store=links
        )

        return self._data
```

tests/mock_module/resource.py
fenekku: discussed IVL -- we are not coupling the two - if ui link changes, it's up to developer to also change it in the links_config

fenekku:
Merging and releasing to get things rolling.

PR #114, opened 2020-09-25
links: support search links [+]

fenekku:


invenio_records_resources/services/records/search.py
```diff
@@ -1,4 +1,13 @@
...
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+# Copyright (C) 2020 Northwestern University.
+#
+# Invenio-Records-Resources is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Search utilities."""
```
ppanero, line 10: Moderate (mergeable if changed). To be consistent with the rest of the `mock_module`.
```suggestion
"""Example search."""
```

In addition @lnielsen we have not set the license header in any file of the `mock_module`. Since it is a mock... Should we?

tests/mock_module/service.py
```diff
@@ -1,12 +1,21 @@
...
+#
+# Invenio-Records-Resources is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
 """Example service."""
 
 from invenio_records_resources.services import RecordService, \
     RecordServiceConfig
+from invenio_records_resources.services.records.search import terms_filter
```
ppanero, line 14: minor: @lnielsen  commented at some point about Elasticsearch DSL `TermsFilter`. However, I cannot find it in their docs now... 🤔

lnielsen:
Integrated and merged as part of #115

invenio_records_resources/services/records/search.py
fenekku: Not in mock_module anymore

Issue #123, opened 2020-10-02
REST API: generate record links for individual records of a search result


fenekku:
Each record of the search result hits, should have their individual links serialized as well.

PR #129, opened 2020-10-14
service: add links to hits of a search result

ppanero:
Closes https://github.com/inveniosoftware/invenio-records-resources/issues/123

Tests added at service level with a mock of the links schema. If the resources define the proper schema links, they will be rendered (tested this manually and is correct).

**Question**: @lnielsen 

Should this task also include the addition of `delete`, `list` etc. links in for the record? Meaning changing `resources/records/schema_links.py` and implementing tests for them?

### DIFF
--- a/tests/services/test_links.py
+++ b/tests/services/test_links.py
@@ -10,52 +10,51 @@
 """Test links."""
 
 import pytest
-from flask_principal import Identity
+from marshmallow import Schema, fields
 
 
-@pytest.fixture()
-def identity_no_need():
-    """Simple identity fixture without needs."""
-    i = Identity(1)
-    return i
+class MockResourceRequestCtx:
+
+    url_args = {"q": ""}
 
 
-@pytest.mark.skip()
-def test_links(service, identity_simple, example_record, es):
+class MockRecordLinksSchema(Schema):
+    """Schema for a record's links."""
+
+    self = fields.String(default="/api/records/12345-ABCDE")
+
+
+class MockSearchLinksSchema(Schema):
+    """Schema for a search result's links."""
+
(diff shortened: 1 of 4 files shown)


tests/services/test_links.py
```diff
@@ -10,52 +10,51 @@
...
 import pytest
-from flask_principal import Identity
+from marshmallow import Schema, fields
 
 
-@pytest.fixture()
-def identity_no_need():
-    """Simple identity fixture without needs."""
-    i = Identity(1)
-    return i
```
fenekku, line 20: We should keep this around to test permission and links even if that tests is skipped for now.

tests/services/test_links.py
```diff
@@ -10,52 +10,51 @@
...
+    """Schema for a search result's links."""
+
+    self = fields.String(default="/api/records?q=")
 
 
-# TODO: example record needs to be permission aware
-@pytest.mark.skip()
-def test_links_with_permissions(service, identity_no_need, example_record, es):
-    """Test record links creation with permissions."""
-    pid_value = str(example_record.id)
```
fenekku, line 52: We should keep this skipped test around: we haven't tested that permission can regulate link generation with the new way of generating links.

fenekku (approved):

ppanero:
@fenekku I was just trying to clean up tests to have as less "skip" as possible. When permissions/access is tackled I assume this part will also be tested. WDYT?

What about opening an issue for it?

ppanero:
issue opened https://github.com/inveniosoftware/invenio-records-resources/issues/133

fenekku:
Thanks for that issue. Merge when you can

PR #135, opened 2020-10-19
search: port bouncing search results solution

### LINKED ISSUE #122: re-enable search preference
Package version (if known): v0.6.2

## Describe the bug

Search preference is not enabled and it could potentially lead to bouncing effect.

## Steps to Reproduce
1. see `services/records/service.py::search()`

## Additional context

It was disabled due to the lack of `user-agent` and `ip` related information. This info should be extracted and set in the `resource_requestctx`, then the resource should pass it to the service.

The old search object uses the `request` object https://github.com/inveniosoftware/invenio-search/blob/81f4d23a1f24fe0d1c2fb7b1a9c64fd7870daed3/invenio_search/api.py#L150 it should be migrated in a way that is not accessed and just passed https://github.com/inveniosoftware/invenio-search/blob/81f4d23a1f24fe0d1c2fb7b1a9c64fd7870daed3/invenio_search/api.py#L270

Values should be accessed at the resource level and passed to the service search? (maybe is the service who calculates the preference (hash) based on what the resource passes.

Why:
- https://github.com/inveniosoftware/invenio-search/pull/107
- https://www.elastic.co/guide/en/elasticsearch/guide/current/_search_options.html#_preference

fenekku:
- :warning: Requires https://github.com/inveniosoftware/invenio-search/pull/207 to be released as 1.4.1
- search: port bouncing search results solution
- tests: clean unused URITemplate import
- close #122

### DIFF
--- a/invenio_records_resources/resources/records/resource.py
+++ b/invenio_records_resources/resources/records/resource.py
@@ -9,7 +9,9 @@
 
 """Invenio Resources module to create REST APIs."""
 
-from flask import abort, g
+import hashlib
+
+from flask import abort, g, request
 from flask_resources import CollectionResource
 from flask_resources.context import resource_requestctx
 
@@ -28,6 +30,15 @@ class RecordResource(CollectionResource, ConfigLoaderMixin):
         super(RecordResource, self).__init__(config=self.load_config(config))
         self.service = service or RecordService()
 
+    def _get_es_preference(self):
+        user_agent = resource_requestctx.headers.get('User-Agent', "")
+        # TODO: Find a way to pass request.remote_addr in resource_requestctx
+        ip = request.remote_addr
+        user_hash = f"{ip}-{user_agent}".encode('utf8')
+        alg = hashlib.md5()
+        alg.update(user_hash)
+        return alg.hexdigest()
+
     #
     # Primary Interface
     #
@@ -38,6 +49,7 @@ class RecordResource(CollectionResource, ConfigLoaderMixin):
             identity=identity,
             params=resource_requestctx.url_args,
             links_config=self.config.links_config,
+            es_preference=self._get_es_preference()
         )
         return hits.to_dict(), 200
 
--- a/invenio_records_resources/services/records/service.py
+++ b/invenio_records_resources/services/records/service.py
@@ -81,7 +81,7 @@ class RecordService(Service):
             )
 
     def create_search(self, identity, record_cls, action='read',
-                      preference=True):
+                      preference=None):
         """Instantiate a search class."""
         permission = self.permission_policy(
             action_name=action, identity=identity)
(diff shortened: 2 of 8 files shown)


lnielsen (approved):

invenio_records_resources/services/records/service.py
```diff
@@ -93,8 +93,7 @@ def create_search(self, identity, record_cls, action='read',
         )
 
         # Avoid query bounce problem
-        if preference:
-            search = search.with_preference_param()
+        search = search.with_preference_param(preference)
 
         # Add document version to ES response
         search = search.params(version=True)
```
ppanero, line 99: Since we have changed operations on invenio-search we can simply chain them now:
```suggestion
        search = search.with_preference_param(preference).params(version=True)
```

invenio_records_resources/resources/records/resource.py
```diff
@@ -28,6 +30,15 @@ def __init__(self, config=None, service=None):
         super(RecordResource, self).__init__(config=self.load_config(config))
         self.service = service or RecordService()
 
+    def _get_es_preference(self):
+        user_agent = resource_requestctx.headers.get('User-Agent', "")
+        # TODO: Find a way to pass request.remote_addr in resource_requestctx
+        ip = request.remote_addr
```
ppanero, line 36: +1 to the TODO 💯 . Otherwise, I don't see the point of having the `resource_requestctx` if we also access `request`.

invenio_records_resources/services/records/service.py
```diff
@@ -126,7 +125,8 @@ def search_request(self, identity, params, record_cls, preference=True):
     #
     # High-level API
     #
-    def search(self, identity, params=None, links_config=None, **kwargs):
+    def search(self, identity, params=None, links_config=None,
+               es_preference=None, **kwargs):
```
ppanero, line 129: Shall we call it `preference`?

invenio_records_resources/services/records/service.py
fenekku: sure

invenio_records_resources/resources/records/resource.py
fenekku: Thanks for reminding me. Made issue here: https://github.com/inveniosoftware/flask-resources/issues/85

invenio_records_resources/services/records/service.py
fenekku: At that interface level (where kwargs can be anything) I preferred keeping the name less generic than "preference" so less chance of future conflict and clearer semantics (it concerns elasticsearch). At that level, knowledge of the underlying search engine is a bit exceptional.

fenekku:
Will rebase with updates after invenio-search https://github.com/inveniosoftware/invenio-search/pull/207 is released.

PR #136, opened 2020-10-20
tests: unskip dynamic config loading

### LINKED ISSUE #92: Dynamic resource/service config loading without app context
The current ConfigLoaderMixin does not allow for running outside the Flask application context because it is accessing the Flask application configuration.

This means you have to do this:

```python
r = Resource(config=ResourceConfig, service=Service(config=ServiceConfig))
app.register_blueprint(r.as_blueprint('records'))
```

but the config is already defined as default config:

```
class Resource:
     default_config = ResourceConfing
class Service:
     default_config = ServiceConfing
```

 Instead, you should just do like this:

```python
r = Resource(service=Service())
app.register_blueprint(r.as_blueprint('records'))
```

fenekku:
closes #92


fenekku:
Merging since of little impact.

PR #137, opened 2020-10-27
config: change hostname var name and add links tests

ppanero:
**Proposal**:
- Move delete, files, etc. links to an issue... Otherwise TODOs will get lost in translation.

### DIFF
--- a/tests/resources/test_resource_links.py
+++ b/tests/resources/test_resource_links.py
@@ -25,14 +25,10 @@ def input_data():
     }
 
 
-def assert_expected_links(pid_value, links):
+def assert_expected_links(pid_value, links, site_hostname="localhost:5000"):
     """Compare generated links to expected links."""
     expected_links = {
-        "self": f"https://localhost:5000/api/mocks/{pid_value}",
-        # TODO: enable delete
-        # "delete": f"https://localhost:5000/api/mocks/{pid_value}",
-        # TODO: enable files
-        # "files": f"https://localhost:5000/api/mocks/{pid_value}/files",
+        "self": f"https://{site_hostname}/api/mocks/{pid_value}",
     }
     assert expected_links == links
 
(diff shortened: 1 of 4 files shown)


tests/resources/test_resource_links.py
```diff
@@ -64,3 +60,19 @@ def test_update_links(app, client, input_data, headers):
 
     assert res.status_code == 200
     assert_expected_links(pid_value, res.json["links"])
+
+
+def test_config_hostname_links(app, client, input_data, headers):
+    import ipdb; ipdb.set_trace()
```
fenekku: Forgotten pdb

ppanero:
closes https://github.com/inveniosoftware/cookiecutter-invenio-rdm/issues/111

tests/resources/test_resource_links.py
```diff
@@ -64,3 +60,18 @@ def test_update_links(app, client, input_data, headers):
 
     assert res.status_code == 200
     assert_expected_links(pid_value, res.json["links"])
+
+
+def test_config_hostname_links(app, client, input_data, headers):
+    site_hostname = "testsite.com"
+    app.config['SITE_HOSTNAME'] = site_hostname
```
fenekku, line 67: moderate: `app` is not a function level fixture, so we have to cleanup after ourselves:

```python
orig_site_hostname = app.config['SITE_HOSTNAME']
app.config['SITE_HOSTNAME'] = site_hostname

# rest of test ...

# cleanup
app.config['SITE_HOSTNAME'] = orig_site_hostname
```

Otherwise we get very strange behaviour.

tests/resources/test_resource_links.py
```diff
@@ -64,3 +60,18 @@ def test_update_links(app, client, input_data, headers):
...
+
+    res = client.post('/mocks', headers=headers, json=input_data)
+    pid_value = res.json["id"]
+    data = res.json
+    data['metadata']['title'] = 'New title'
+
+    res = client.get(f'/mocks/{pid_value}', headers=headers, json=data)
```
fenekku, line 74: We can remove this for the purpose of this test. The API response to POST should include links that have the app's SITE_HOSTNAME and can be used in the assertion directly.

fenekku (approved):
Good after changes.

tests/resources/test_resource_links.py
ppanero: Nice catch!

tests/resources/test_resource_links.py
ppanero: Indeed, thanks!

PR #141, opened 2020-11-10
records-models: record file base classes

### LINKED ISSUE #143: files [models]: record files management
For connecting records to files we need a new set of models/fields that manage:

- Listing of a record/dratft's files
- State of the file upload in the draft (initialized, pending, completed)
- Storing metadata for the files (e.g. "description", "width/height", etc.)

For that purpose we need the following new "base/mixin" models:

```python
# NOTE: being a sub-class RecordMetadata allows storing file metadata in `json`
class RecordFileBase(invenio_records.models.RecordMetadataBase):

    key = db.Column(db.Text())
    """Filename (could be also a long path)."""

    record_model_cls = None
    """Used to specify the record model fro the FK."""

    record_id = db.Column(
        UUIDType,
        db.ForeignKey(cls.record_model_cls.id),
        nullable=False,
    )
    record_id = db.relationship(cls.record_model_cls)

    object_version_id = db.Column(
        UUIDType,
        db.ForeignKey(ObjectVersion.id),
        nullable=True,  # filled-in when the upload is finalized
    )
    object_version = db.relationship(ObjectVersion)

    # To ensure unique filenames in a record
    __table_args__ = (
        db.UniqueConstraint('id', 'key'),
    )


# We extend the base record class with a bucket
class RecordMetadataBase(invenio_records.models.RecordMetadataBase):

    bucket_id = db.Column(
        UUIDType,
        db.ForeignKey(Bucket.id),
        nullable=True,
    )
    bucket = db.relationship(Bucket)
```

slint:
Closes #143.

### DIFF
--- a/invenio_records_resources/records/models.py
+++ b/invenio_records_resources/records/models.py
@@ -1,0 +1,67 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+#
+# Invenio-Records-Resources is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Records Models."""
+
+import invenio_records.models as base_models
+from invenio_db import db
+from invenio_files_rest.models import ObjectVersion
+from sqlalchemy.dialects import mysql
+from sqlalchemy.ext.declarative import declared_attr
+from sqlalchemy_utils.types import UUIDType
+
+
+class RecordFileBase(base_models.RecordMetadataBase):
+    """Base class for a record file, storing its state and metadata."""
+
+    record_model_cls = None
+    """Record model to be for the ``record_id`` foreign key."""
+
+    key = db.Column(
+        db.Text().with_variant(mysql.VARCHAR(255), 'mysql'),
+        nullable=False,
+    )
+    """Filename key (can be path-like also)."""
+
+    @declared_attr
+    def record_id(cls):
+        """Record ID foreign key."""
+        return db.Column(
+            UUIDType,
+            db.ForeignKey(cls.record_model_cls.id, ondelete='RESTRICT'),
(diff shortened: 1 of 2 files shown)


invenio_records_resources/records/models.py
```diff
@@ -0,0 +1,85 @@
...
+                'id', 'key', unique=True,
+            ),
+        )
+
+
+class RecordMetadataBase(base_models.RecordMetadataBase):
+    """Represent a base class for record metadata with files."""
+
+    @declared_attr
+    def bucket_id(cls):
```
ppanero, line 74: comment: if we add a `bucket` column on the db level, we are assuming there is no chance of using records without files (e.g. someone implementing their app using only metadata). Or at least they will have an empty column in every row. This is fine for me, even more taking into account the time constraint. However, given this, what we discussed at service level could be made tight with files (still with mixins) but without taking into account the no files case.

ppanero (commented):
LGTM

invenio_records_resources/records/models.py
```diff
@@ -0,0 +1,85 @@
...
+from sqlalchemy.dialects import mysql
+from sqlalchemy.ext.declarative import declared_attr
+from sqlalchemy_utils.types import UUIDType
+
+
+class RecordFileBase(base_models.RecordMetadataBase):
+    """Base class for a record file, storing its state and metadata."""
+
+    record_model_cls = None
+    """."""
```
ppanero, line 23: minor: a minimal description instead of a dot ;) :P

invenio_records_resources/records/models.py
slint: I agree. I've removed it from the default classes and added it only in tests. This will be added concretely in `invenio-rdm-records`.

slint:
Included in #154

Issue #142, opened 2020-11-16
files [resources]: Implement file resources at record level


ppanero:
[CodiMD source](https://codimd.web.cern.ch/N8p4gXDOROiPTutnQfkzsg?view)

**Important note**: At this level, *records-resources*, all actions are implemented to support files when records do not need drafts. However, the endpoints will be overwritten at *drafts-resources* level.

**/records/:id/files**
- Type: Collection resource (list view)
- Actions:

| *Verb*         |                                        |                                     |
|:----------------- |:-------------------------------------- |:----------------------------------- |
| GET               | /records/:id/files                     | List files metadata                 |
| POST              | /records/:id/files               | Initialize file upload     |
| PUT               | /records/:id/files        | Update metadata of files. TBD. **Do not implement**             |
| DELETE            | /records/:id/files               | Delete all files                    |

**/records/:id/files/:key**
- Type: Resource (item view of the collection above)
- Actions:

| *Verb*         |                                        |                                     |
|:----------------- |:-------------------------------------- |:----------------------------------- |
| PUT               | /records/:id/files/:key          | Update metadata of 1 file           |
| GET               | /records/:id/files/:key          | Get 1 file metadata       |
| DELETE            | /records/:id/files/:key          | Delete 1 file [2]                       |

*Note* Complete download passed from `POST: /records/:id/files/:key` to be a `complete` action.

**/records/:id/files/:action**
- Type: Action resource (Using singleton resource)
- With the following actions configured:

| *Verb*         |                                        |                                     |
|:----------------- |:-------------------------------------- |:----------------------------------- |
| PUT               | /records/:id/files/:key/content [1] | File upload if directly via Invenio |
| GET               | /records/:id/files/:key/content [1] | Download file             |
| POST              | /records/:id/files/:key/commit         | Complete the file upload            |

This will enable to support different actions in the future (e.g. processes like transcoding, compress, etc)

PR #152, opened 2020-11-16
resources: implement base action and files

ppanero:
Closes https://github.com/inveniosoftware/invenio-records-resources/issues/142

**Question**

- Shall we put actions to prefix content {down/up}load and commit? i.e. `.../:key/actions/<action>`. This is the pattern use for example for `publish`ing a draft. But I think it's more UX friendly without it.

**IMPORTANT NOTE**
Note that content upload/download and commit will return a 202 (as actions), not a 200. Changing this might require further re-write of the action resource, can be done in the future as a separate task (even outsource since it should not be too complicated).

**IMPORTANT NOTE TOO**

I had to rewrite a bit the actions to be able to reuse what was done in `invenio-drafts-resources` for publishing. It is not optimal (... time constraints ... meh). But it will allow a creation of actions with one line (only define config) in `invenio-drafts-resources` in the same fashion that is done here for files. Otherwise it was copy-paste code 3 times (files here, files in drafts and in drafts itself.)


invenio_records_resources/resources/actions/resource.py
```diff
@@ -0,0 +1,82 @@
...
+        return item.to_dict(), 202
+
+    # SingletonView GET
+    def read(self, *args, **kwargs):
+        """GET operations on actions."""
+        item = self.handle_action_request('read')
+
+        return item.to_dict(), 202
+
+    # SingletonView DETELE
```
slint, line 77:
```suggestion
    # SingletonView DELETE
```

invenio_records_resources/resources/actions/resource.py
```diff
@@ -0,0 +1,82 @@
...
+    def delete(self, *args, **kwargs):
+        """GET operations on actions."""
```
slint, line 79:
```suggestion
        """DELETE operations on actions."""
```

slint (approved):
Some *important* typos, otherwise LGTM

PR #173, opened 2020-12-07
services: allow to create w/o raising errors only reporting

fenekku:
- also starts December release series 0.9.X
- part of closing https://github.com/inveniosoftware/invenio-drafts-resources/issues/81

### DIFF
--- a/invenio_records_resources/services/records/service.py
+++ b/invenio_records_resources/services/records/service.py
@@ -170,16 +170,27 @@ class RecordService(Service):
         return self._create(
             self.record_cls, identity, data, links_config=links_config)
 
-    def _create(self, record_cls, identity, data, links_config=None):
+    def _create(self, record_cls, identity, data, links_config=None,
+                raise_errors=True):
         """Create a record.
 
         :param identity: Identity of user creating the record.
-        :param data: Input data according to the data schema.
+        :param dict data: Input data according to the data schema.
+        :param links_config: Links configuration.
+        :param bool raise_errors: raise schema ValidationError or not.
         """
         self.require_permission(identity, "create")
+        # partial = partial or tuple()
 
         # Validate data and create record with pid
-        data, _ = self.schema.load(identity, data)
+        data, errors = self.schema.load(
+            identity,
+            data,
+            raise_errors=raise_errors  # if False, flow is continued with data
+                                       # only containing valid data, but errors
+                                       # are reported (as warnings)
+        )
+
         # It's the components who saves the actual data in the record.
         record = record_cls.create({})
 
@@ -198,7 +209,8 @@ class RecordService(Service):
             self,
             identity,
             record,
-            links_config=links_config
+            links_config=links_config,
+            errors=errors
         )
 
     def read(self, id_, identity, links_config=None):
--- a/invenio_records_resources/version.py
+++ b/invenio_records_resources/version.py
@@ -13,4 +13,4 @@ This file is imported by ``invenio_records_resources.__init__``,
 and parsed by ``setup.py``.
 """
 
-__version__ = "0.8.8"
+__version__ = "0.9.0"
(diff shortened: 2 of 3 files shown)


invenio_records_resources/services/records/service.py
```diff
@@ -170,16 +170,27 @@ def create(self, identity, data, links_config=None):
...
 
-    def _create(self, record_cls, identity, data, links_config=None):
+    def _create(self, record_cls, identity, data, links_config=None,
+                raise_errors=True):
         """Create a record.
 
         :param identity: Identity of user creating the record.
         :param data: Input data according to the data schema.
+        :param links_config: Links configuration.
+        :param partial: Marshmallow 'partial' argument.
```
ppanero, line 180: minor: partial is not used anywhere (nor accepted by the function definition).

invenio_records_resources/version.py
```diff
@@ -13,4 +13,4 @@
 and parsed by ``setup.py``.
 """
 
-__version__ = "0.8.8"
+__version__ = "0.9.0"
```
ppanero, line 16: comment: talk maybe to @zzacharo or @kprzerwa because they are doing some work on this module as part of the sprint, just to make sure it does not collide.

tests/services/test_service__create.py
```diff
@@ -0,0 +1,51 @@
+# -*- coding: utf-8 -*-
```
ppanero, line 1: moderate: double underscore in the file name. `test_service__create.py` --> `test_service_create.py`

ppanero:
Leaving the approval to @zzacharo but dont wait If you are blocked @fenekku. I just wanna make sure things don't collide :)

invenio_records_resources/version.py
zzacharo: @fenekku from my side I thin is fine but let's talk once you are online again! Btw, invenio-app-rdm has already released in `0.9.0`.

zzacharo (approved):

invenio_records_resources/services/records/service.py
```diff
@@ -198,7 +209,8 @@ def _create(self, record_cls, identity, data, links_config=None):
             self,
             identity,
             record,
-            links_config=links_config
+            links_config=links_config,
+            errors=errors
```
zzacharo, line 213: comment: Have you discussed how the UI will show errors/warnings if they are coming from saving a draft? I understand that in this case the save will be complete but still, errors will be returned! Will we have a different way of showing these warnings? If that's the case, why don't we differentiate them already here in "warnings" or something that indicates that these errors at that point are not fatal for the UI...

zzacharo (approved):

invenio_records_resources/services/records/service.py
slint: I think the overall idea is that for updates (i.e. `PUT /api/records/:id/draft`), we basically always give back a 200, with an `errors` key. For publishing though (i.e. `POST /api/records/:id/draft/actions/publish`), we give back a 400 in case of validation errors... The question is, should we keep the response payload homogenous, i.e. still return both the partial valid record, and the `errors` key at the top level?

slint: > ...we basically always give back a 200, with an `errors` key (if there are errors)...

There are some "fancier" ways of communicating a partial success, via the [`207 Multi-Status` response](https://evertpot.com/http/207-multi-status), but since there's nothing "standard", probably let's stick with what we know :)

invenio_records_resources/services/records/service.py
fenekku: Thanks! That's a leftover.

tests/services/test_service__create.py
fenekku: yeah.. because we are testing `_create` and not `create` :sweat_smile:

invenio_records_resources/services/records/service.py
fenekku: yeah I was thinking of sticking to 200 / 400 and then like @slint mentioned: because we know the context in which we are getting a payload back with errors (following a save draft or a publish), the frontend can use that to tweak the display. Save draft would display with `warning` (orange) and publish with `error` (red).

PR #190, opened 2021-01-20
systemfields: factors out create/delete from hooks

lnielsen:
* Moves create/delete pid code out of the life-cycle hooks so it can be
  called outside the life-cycle hooks.

### DIFF
--- a/invenio_records_resources/records/systemfields/pid.py
+++ b/invenio_records_resources/records/systemfields/pid.py
@@ -106,33 +106,42 @@ class PIDField(RelatedModelField):
             context_cls=context_cls,
         )
 
-    #
-    # Life-cycle hooks
-    #
-    def post_create(self, record):
-        """Called after a record is created."""
-        if self._provider is None or not self._create:
-            return
-
-        # This uses the data descriptor method __get__() below:
-        if getattr(record, self.attr_name) is None:
+    def create(self, record):
+        """Method to create a new persistent identifier for the record."""
+        # This uses the fields __get__() data descriptor method below
+        pid = getattr(record, self.attr_name)
+        if pid is None:
             # Create a PID if the object doesn't already have one.
-            _pid = self._provider.create(
+            pid = self._provider.create(
                 object_type=self._object_type,
                 object_uuid=record.id,
                 record=record,
             ).pid
 
-            setattr(record, self.attr_name, _pid)
+            # Set using the __set__() method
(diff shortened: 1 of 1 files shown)


invenio_records_resources/records/systemfields/pid.py
```diff
@@ -106,33 +106,42 @@ def __init__(self, key='id', provider=None, pid_type=None,
...
-        if self._provider is None or not self._create:
-            return
-
-        # This uses the data descriptor method __get__() below:
-        if getattr(record, self.attr_name) is None:
+    def create(self, record):
+        """Method to create a new persistent identifier for the record."""
+        # This uses the fields __get__() data descriptor method below
+        pid = getattr(record, self.attr_name)
+        if pid is None:
```
ppanero, line 113: Minor: It assumes `attr_name` always exists? I think is a fair assumtion since it is assigned in the `SystemField` constructor. However, if not, it will throw an exception...

```
>>> class Test:
...     exists = "yes"
...
>>> getattr(Test, "exists")
'yes'
>>> getattr(Test, "nope")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: type object 'Test' has no attribute 'nope'
```

this means that it will fail if it does not exists, it should be: `pid = getattr(record, self.attr_name, None)`, then is harmonized to return None

ppanero (approved):
Out of curiosity: In which case do they need to be created/deleted outside of the workflow?

lnielsen:
> Out of curiosity: In which case do they need to be created/deleted outside of the workflow?

It's needed when the PID is dependent on the provided data. In the service ``create()``, we do something like:

```python
data = {'id': 'eng', ...}
record = Record.create({})
record.update(data)
```

Above, I want a ``PersistentIdentifier(pid_value=record['id'], ...)``  but because of the empty record being created, then there's no data in ``record['id']``. So instead the workflow is now:

```python
data = {'id': 'eng', ...}
record = Record.create({})
record.update(data)
record.pid.create()
```

invenio_records_resources/records/systemfields/pid.py
lnielsen: I would leave it as-is, because it's really a big mistake if ``attr_name`` is not set. The property is set automatically when the system field (i.e. data descriptor) is attached to a class via ``__set_name__()``, and not having it means that things should break and not work.

PR #191, opened 2021-01-21
service: add read_many and read_all methods

### LINKED ISSUE #189: services: add new high-level api for reading many records at once
For Vocabularies, we have two use cases:

- Retrieve a lower number of records (~10) that are being displayed in facets (e.g. 10 language records)
  - For instance language vocabulary has 1000+ records so unfeasible to fetch all.
- Retrieve all records for a given type (<100) and potentially cache the result under a provided key
  - For instance, resource type has ~20 records, and all can be cached so when facets are displayed we just retrieve from the cache.

The goal here is that the methods are very performant as they will happen during search.

**Read many**

The API of read many *could* (but doesn't have to) look like this:

```
service.read_many(identity, ids=['', ...])
```

It could use the [search engine multi get](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-multi-get.html). For that we require uuids to be provided, so either a query must be made to PIDStore to retrieve all uuids. Alternatively we get everthing from the DB.

**Read all**

Perhaps the ``scan()`` method can be extended with cache capabilities or a ``read_all()`` method simply uses the same as ``read_many()`` just retrieving everything.

```
service.scan(identity, params={}, cache_key='vocabs.resource_types')
```

E.g. in vocabularies, we would need to query by the type for generic vocabularies.

ppanero:
closes #189 
requires https://github.com/inveniosoftware/invenio-records/pull/254

Issues/questions:

- Is this gonna be used at resource level? The `GET` methods are already used, so we would need new URLs (an issue at RDM-records level).
- Cache seems to not be cleaned from test to test, but there is no fixture to do so.
- Elasticsearch DSL Response is not pickable, therefore is not cachable:

```
ipdb> test = search.execute()
ipdb> test
<Response: [<Hit(records-record-v1.0.0-1611231064/93511d5c-fc20-4163-b48f-3c8fa46f3959): {'$schema': 'http://localhost/schemas/records/record-v1.0.0....}>, <Hit(records-record-v1.0.0-1611231064/cb7af96d-5991-4e34-a516-090f8bcaf8cf): {'$schema': 'http://localhost/schemas/records/record-v1.0.0....}>]>
ipdb> current_cache.set("test", test)
True
ipdb> current_cache.get("test")
*** RecursionError: maximum recursion depth exceeded while calling a Python object
```

- An alternative is to cache the `to_dict` of the result_list, but then in some cases the service returns a `ResultList` and in others a `dict`... This would be solved by having a `from_dict` method at `ResultList` level, but is not trivial to implement (Since most of its properties assume an ES `Response` as `_records` not a `Dict`.


ppanero:
@lnielsen closing this since I guess is old

lnielsen (commented):
Discuss IRL.....extract only fields.

ppanero:
From IRL:

- fields support from when?
   Cannot use it. Comes in 7.10: [Add search fields option to support high-level field retrieval #60100 (issues: #49028, #55363)](https://www.elastic.co/guide/en/elasticsearch/reference/current/release-notes-7.10.0.html)
- bare schema works VS create a new result type
- spit out same as a 'records list'
- key -> id_key
- filter_fields mandatory (diff than the key), if None... Do not source filter
- extract key to the config, it's the pid of the record (record.pid.pid_value, assume it exists)
- `service.read_all(identity, fields=['title.en'], cache=True)`
- `service.read_many(identity, ['eng', 'dan'], fields=['title.en'])`
- source filtering optional (return entire record) --> `service.read_many(identity, ['eng', 'dan'])

ppanero:
@lnielsen this is ready for review... is failing now on due to Werkzeug...

ppanero:
local tests are giving me:
```

================================================================================================== short test summary info ===================================================================================================
FAILED tests/services/test_service.py::test_read_all_cache - AttributeError: 'Response' object has no attribute 'labelled_facets'
```

invenio_records_resources/services/records/service.py
```diff
@@ -288,6 +290,72 @@ def read(self, id_, identity):
...
+            permission_action='search',
+        )
+
+        # Fetch only certain fields - explicitly add internal system fields
+        # required to use the result list to dump the output.
+        if fields:
+            dumper_fields = [
+                "uuid", "version_id", "created", "updated", "expires_at"]
+            fields = fields + dumper_fields
+            # ES 7.11+ supports a more efficient way of fetching only certain
```
lnielsen, line 311: FYI: I've rebased and made some minor edits of notes.

invenio_records_resources/services/records/service.py
```diff
@@ -288,6 +290,72 @@ def read(self, id_, identity):
...
+            # ES 7 versions does not support it, so we use the source filtering
+            # method instead for now.
+            search = search.source(fields)
+
+        search_result = search.query(es_query).execute()
+
+        return search_result
+
+    def read_many(self, identity, ids, fields=None,
+                  record_cls=None, search_opts=None, **kwargs):
```
lnielsen, line 322: The high-level service methods ``read_many()`` and ``read_all()`` should not take ``record_cls`` and ``search_opts`` as inputs (it's ok for ``_read_many()`` to do so). Rather, if we need for drafts we would need to add corresponding ``read_all_drafts()`` and ``read_many_drafts()`` methods (however I dont' think we need them, since it's mainly for vocabularies we need them.

invenio_records_resources/services/records/service.py
```diff
@@ -288,6 +290,72 @@ def read(self, id_, identity):
...
+                fields=fields,
+                record_cls=record_cls,
+                search_opts=search_opts,
+                **kwargs
+            )
+
+            results = self.result_list(self, identity, results)
+
+            if cache:
+                # NOTE: ES DSL Response is not pickable.
```
lnielsen, line 353: FYI: Prefixing a comment with ``# NOTE:`` has a special meaning, when it's above a translation text, e.g.:

```
# NOTE: This is a form-label displayed to the user
label = _('....')
```

This will extract the note and put it in the translation files for translators to better understand what's going on, hence, it's better to keep the ``# NOTE:`` prefixing for only those siuations and simply not prefix at all.

tests/services/test_service.py
```diff
@@ -70,3 +71,99 @@ def test_simple_flow(app, consumer, service, identity_simple, input_data):
...
+    records = service.read_all(
+        identity_simple, fields=["metadata.title"])  # cached by default
+
+    assert records.total == 2
+    for record in records.hits:
+        assert record["id"] is not None
+        assert list(record["metadata"].keys()) == ["title"]
+        assert record["metadata"]["title"] == "Test"
+
+    cached = current_cache.get("metadata.title")
```
lnielsen, line 93: Minor: Here I would call again ``read_all()`` to ensure that it works. I think you can just assert that there's a cache key in the cache without testing what is being cached.

tests/services/test_service.py
```diff
@@ -70,3 +71,99 @@ def test_simple_flow(app, consumer, service, identity_simple, input_data):
     # - search
     res = service.search(identity_simple, q=f"id:{id_}", size=25, page=1)
     assert res.total == 0
+
```
lnielsen, line 74: Perhaps a fixture like this would work to avoid having to call ``current_cache.clear()`` at both the start and end of the tests.

```python
@pytest.fixture()
def cache():
    try:
        yield current_cache
    finally:
        current_cache.clear()
```

invenio_records_resources/services/records/service.py
ppanero: oops didn't know that... gonna have to change some places 😓

invenio_records_resources/services/records/service.py
ppanero: makes sense, thanks :)

tests/services/test_service.py
ppanero: Fair, I wanted to sort of make sure that the cache was correct... but is true that its gonna be the same that gets returned...

tests/services/test_service.py
ppanero: Yeah... wanted to do that in `pytests-invenio`.... but I'll do it here and move there later...

invenio_records_resources/services/records/service.py
```diff
@@ -288,6 +291,71 @@ def read(self, id_, identity):
...
+
+        else:
+            search = self.create_search(
+                identity=identity,
+                record_cls=self.record_cls,
+                search_opts=self.config.search,
+                permission_action='search',
+            ).query(es_query)
+
+            results = Response(search, results)
```
ppanero, line 355: Need to reconstruct the ES response, so the `result_list` item works out of the box 🚀

ppanero:
@lnielsen note there is a release commit also included ;)

invenio_records_resources/services/records/service.py
```diff
@@ -288,6 +291,71 @@ def read(self, id_, identity):
...
+        clauses = []
+        for id_ in ids:
+            clauses.append(Q('term', **{"id": id_}))
+        query = Q('bool', minimum_should_match=1, should=clauses)
+
+        results = self._read_many(identity, query, fields, **kwargs)
+
+        return self.result_list(self, identity, results)
+
+    def read_all(self, identity, fields, cache=True, **kwargs):
```
fenekku, line 333: Just another wrinkle I ran into while producing the resource type choices on the upload page:

There we need, for each resource type:

- id
- type name
- subtype name

type name and subtype name are found in the "props" . I am not sure how we want to deal with those. For now, in invenio-app-rdm I generate the vocabulary list from the database. :shrug:

ppanero: would passing `props.type` and `props.subtype` in the `fields` parameter work? sort of like [here](https://github.com/inveniosoftware/invenio-records-resources/pull/191/files#diff-0be63bb442d15c429661d7b94b34276b9bae5cf6c81562f2699b78e3bdfafe55R77)

lnielsen (approved):

PR #198, opened 2021-02-17
service: add option to override the index on the service level

zzacharo:


invenio_records_resources/records/systemfields/has_draftcheck.py
```diff
@@ -0,0 +1,52 @@
...
+        :param key: Attribute name of the PIDField to use for status check.
+        :param status: The status or list of statuses which will return true.
+        """
+        # assert status, "You must provide a PIDStatus to test against."
+        self._draft_cls = draft_cls
+
+    #
+    # Data descriptor methods (i.e. attribute access)
+    #
+    def __get__(self, record, owner=None):
```
lnielsen, line 44: For the logic on caching - see https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/systemfields/relatedmodelfield.py#L88-L139

invenio_records_resources/services/records/service.py
```diff
@@ -252,6 +252,7 @@ def read(self, id_, identity, links_config=None):
         """Retrieve a record."""
         # Resolve and require permission
         # TODO must handle delete records and tombstone pages
+        import ipdb; ipdb.set_trace()
```
lnielsen, line 255: ipdb

PR #208, opened 2021-03-03
650 record item permissions for

fenekku:
- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/650
- releases it so that invenio-app-rdm can use it

### DIFF
--- a/invenio_records_resources/services/records/results.py
+++ b/invenio_records_resources/services/records/results.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2020 CERN.
-# Copyright (C) 2020 Northwestern University.
+# Copyright (C) 2020-2021 CERN.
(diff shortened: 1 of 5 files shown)


invenio_records_resources/services/records/results.py
```diff
@@ -70,6 +70,32 @@ def to_dict(self):
             res['errors'] = self._errors
         return res
 
+    def permissions_to(self, actions):
```
zzacharo, line 73: I would probably make the name a bit more explicit to something like `has_permission_to(...)`

zzacharo:
LGTM after comment is addressed :)

invenio_records_resources/services/records/results.py
fenekku: Sure. I will go for `has_permissions_to(...)` (keeping the plural because multiple permissions).

PR #215, opened 2021-03-22
services: specify record kwarg for schema.dump in ResultItem

max-moser:
closes inveniosoftware/invenio-rdm-records#439
in `ResultItem.data`, the `record` keyword argument was missing from the call to `self._service.schema.dump()`, causing the `record` argument in the permission generators to be set to `None`

### DIFF
--- a/invenio_records_resources/services/records/results.py
+++ b/invenio_records_resources/services/records/results.py
@@ -57,6 +57,7 @@ class RecordItem(ServiceItemResult):
         self._data = self._service.schema.dump(
             self._identity,
             self._record,
+            record=self._record,
             links_namespace="record",
             links_factory=links,
         )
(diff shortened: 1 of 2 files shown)


invenio_records_resources/services/records/results.py
```diff
@@ -57,6 +57,7 @@ def data(self):
         self._data = self._service.schema.dump(
             self._identity,
             self._record,
+            record=self._record,
```
fenekku, line 60: Good catch. Honestly I think a refactor of `MarshmallowServiceSchema` / `FieldPermissionsMixin` (marshmallow-utils) is in order to take a context parameter and have the record itself be taken into account.

fenekku (approved):

Issue #225, opened 2021-04-09
files: move record.files.enabled to Record API


lnielsen:
First perform the changes in Invenio-Drafts-Resources - this will enable the new API to be used. Next, make changes to React-Invenio-Deposit, to use the new API. Last, remove the options from the previous API.

### In Invenio-Drafts-Resources:

- DraftFilesComponent: Implement the logic previously placed in  ``update_files_options()``

### In Invenio-Records-Resources

- FileResource:
   - Remove ``update_all()``
- FileSerivce:
  - Remove ``update_files_options()``

### React-Invenio-Deposit

- To be better specified exactly what's needed.

### Docs

- Update REST API description
  -  https://inveniordm.docs.cern.ch/reference/rest_api/#modify-a-drafts-files-options
  - Add new options to the record.

PR #229, opened 2021-04-12
files: enable/disable files via record body

fenekku:
- part of closing #225
  * defines component to assign `files.enabled`
  * ended up having to do the majority of the changes here because file addition is defined in this package
- renames `tests/mock_module/service.py` -> `tests/mock_module/config.py`
- removes `unknown = INCLUDE` in keeping with previous work (see https://github.com/inveniosoftware/invenio-rdm-records/issues/496)
- moves some fixtures up the hierarchy so they can be reused

### DIFF
--- a/invenio_records_resources/services/files/service.py
+++ b/invenio_records_resources/services/files/service.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2020 CERN.
-# Copyright (C) 2020 Northwestern University.
+# Copyright (C) 2020-2021 CERN.
+# Copyright (C) 2020-2021 Northwestern University.
 #
 # Invenio-Records-Resources is free software; you can redistribute it and/or
 # modify it under the terms of the MIT License; see LICENSE file for more
@@ -103,10 +103,6 @@ class FileService(Service):
         action = self.config.permission_action_prefix + "update_files"
         self.require_permission(identity, action, record=record)
 
-        # TODO: Maybe there's a better programmatic API to apply these?
-        # e.g. record.files.update(...)
-        if data.get('enabled') is not None:
-            record.files.enabled = data['enabled']
         if record.files.enabled:
             if 'default_preview' in data:
                 record.files.default_preview = data['default_preview']
--- a/invenio_records_resources/services/records/components.py
+++ b/invenio_records_resources/services/records/components.py
@@ -1,12 +1,16 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2020 CERN.
+# Copyright (C) 2020-2021 CERN.
+# Copyright (C) 2021 Northwestern University.
 #
 # Invenio-Records-Resources is free software; you can redistribute it and/or
 # modify it under the terms of the MIT License; see LICENSE file for more
 # details.
 
 """Records service component base classes."""
+
+from flask_babelex import gettext as _
+from marshmallow import ValidationError
 
 
 class ServiceComponent:
@@ -72,3 +76,41 @@ class MetadataComponent(ServiceComponent):
     def update(self, identity, data=None, record=None, **kwargs):
         """Inject parsed metadata to the record."""
         record.metadata = data.get('metadata', {})
+
+
+class FilesOptionsComponent(ServiceComponent):
+    """Service component for files' options.
+
+    It only deals with enabled / disabled (metadata-only) files for now.
+    """
+
+    def _validate_files_enabled(self, record, enabled):
+        """Validate files enabled."""
+        if not enabled:
+            if record.files.values():
+                raise ValidationError(
+                    _("You must first delete all files to set the record to "
+                      "be metadata-only."),
+                    field_name="files.enabled"
+                )
+
+    def assign_files_enabled(self, identity, enabled, record=None, **kwargs):
+        """Assign files enabled.
+
+        This is a public interface so that it can be reused elsewhere
+        (e.g. drafts-resources).
+        """
+        self._validate_files_enabled(record, enabled)
+        record.files.enabled = enabled
+
+    def create(self, identity, data=None, record=None, **kwargs):
+        """Inject parsed files options in the record."""
+        # presence is guaranteed by schema
+        enabled = data["files"]["enabled"]
+        self.assign_files_enabled(identity, enabled, record, **kwargs)
+
+    def update(self, identity, data=None, record=None, **kwargs):
+        """Inject parsed files options in the record."""
+        # presence is guaranteed by schema
+        enabled = data["files"]["enabled"]
+        self.assign_files_enabled(identity, enabled, record, **kwargs)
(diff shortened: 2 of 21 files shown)


invenio_records_resources/services/records/schema.py
```diff
@@ -53,12 +41,42 @@ def clean(self, data, **kwargs):
         return data
 
 
+class TypeSchema(Schema):
```
lnielsen, line 44: Question: What's the reason for adding the type schema? As this is data model specific I would not add it here.

I aware the others below should like ``metadata.title`` should likely not be here either.

invenio_records_resources/services/records/components.py
```diff
@@ -72,3 +51,40 @@ def create(self, identity, data=None, record=None, **kwargs):
...
+
+    It only deals with enabled / disabled (metadata-only) files for now.
+    """
+
+    def _validate_files_enabled(self, record, enabled):
+        """Validate files enabled."""
+        if not enabled:
+            if record.files.values():
+                raise ValidationError(
+                    _("Can't set record with files to be metadata-only."),
```
lnielsen, line 67:
```suggestion
                    _("You must first delete all files to set the record to be metadata-only."),
```

invenio_records_resources/services/records/components.py
```diff
@@ -72,3 +51,40 @@ def create(self, identity, data=None, record=None, **kwargs):
     def update(self, identity, data=None, record=None, **kwargs):
         """Inject parsed metadata to the record."""
         record.metadata = data.get('metadata', {})
+
+
+class FilesOptionsComponent(ServiceComponent):
```
lnielsen, line 56: The component should also deal with setting ``record.files.preview`` and ``record.files.order``.

tests/resources/test_files_resource.py
```diff
@@ -385,3 +392,40 @@ def test_disabled_upload_file_resource(client, es_clear, headers, location):
         f"/mocks_disabled_files_upload/{id_}/files/test.pdf", headers=headers
     )
     assert res.status_code == 405
+
+
+def test_disable_files_when_files_already_present_should_error(
+        app, client, es_clear, headers, input_data, location):
+    # Initialize a record
+    response = client.post('/mocks', headers=headers, json=input_data)
```
lnielsen, line 400: Question: is it always required to send ``input_data`` now?

invenio_records_resources/services/records/schema.py
```diff
@@ -20,17 +19,6 @@
 #
 # The default record schema
 #
-class MetadataSchema(Schema):
-    """Basic metadata schema class."""
-
-    class Meta:
-        """Meta class to accept unknown fields."""
-
-        unknown = INCLUDE
```
lnielsen, line 29: 👍

invenio_records_resources/services/records/schema.py
```diff
@@ -53,12 +41,42 @@ def clean(self, data, **kwargs):
...
+          recommended.
+    """
 
     metadata = fields.Nested(MetadataSchema)
 
 
+class FilesOptionsSchema(Schema):
+    """Basic files options schema class."""
+
+    enabled = fields.Bool(missing=True)
```
lnielsen, line 71: See comment on FilesComponent related to ``order`` and  ``preview``.

invenio_records_resources/services/records/components.py
lnielsen, line 56: Major: Let's discuss this during a call today. I think this should be in DraftFilesComponent in Invenio-Drafts-Resources.

lnielsen (commented):
All the minor restructuring looks good. The FilesComponent location is the main issue we'll should discuss today in a call.

invenio_records_resources/services/files/service.py
```diff
@@ -103,10 +103,6 @@ def update_files_options(self, id_, identity, data):
         action = self.config.permission_action_prefix + "update_files"
         self.require_permission(identity, action, record=record)
 
-        # TODO: Maybe there's a better programmatic API to apply these?
```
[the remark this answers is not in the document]
fenekku: Eventually maybe? I favored starting small and targeted to the issue at hand. Because I was already introducing a lot of changes, I went on the conservative side (these options might also have large ramifications). I made an issue to transition them too: https://github.com/inveniosoftware/invenio-records-resources/issues/230

invenio_records_resources/services/records/components.py
```diff
@@ -37,31 +41,6 @@ def search(self, identity, search, params, **kwargs):
         return search
 
 
-class DataComponent(ServiceComponent):
```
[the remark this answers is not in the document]
fenekku: Ah my bad! I was checking app-rdm, rdm-records and drafts-resources, but missed invenio-vocabularies. Thanks, re-instating!

invenio_records_resources/services/records/components.py
fenekku: I saw this in the original ticket and it was mentioned during the call yesterday, but since all of the file logic/responsibility was in here, I went for implementing it here to be reused in drafts-resources. But if we do move *all* the file logic out of records-resources and into drafts-resources then that should work. Otherwise, records-resources still needs to implement files options i.e. doesn't provide everything needed to create records with files.

invenio_records_resources/services/records/schema.py
fenekku: The faceting tests in records-resources index a `type` field with children `type` + `subtype`. By removing the `INCLUDE` above, those needed to be added explicitly. I do agree that in fact, just `BaseRecordSchema` should be defined in `schema.py` and that tests (`mock module`) should be the one defining the `RecordSchema` with title + type. If that makes sense, I can do that change?

tests/resources/test_files_resource.py
```diff
@@ -222,6 +227,7 @@ def test_default_preview_file(app, client, es_clear, headers, location):
     assert res.json['default_preview'] is None
 
 
+@pytest.mark.skip()
```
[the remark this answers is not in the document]
fenekku: Yes agreed :+1:

tests/resources/test_files_resource.py
fenekku: For the `RecordWithFilesSchema`, the `files` attribute is required so `input_data` is required (with `files` filled). It could be made optional... It introduced some questions I wasn't sure how to answer: default to True probably? but at the component level or at the schema level? and where can we assume that we have a value for that field - don't want to always be checking.... Forcing `required` was explicit :sweat_smile: . Not attached to any approach though.

invenio_records_resources/services/records/schema.py
fenekku: Did the change :nerd_face:

fenekku:
We went over this in video chat and I removed the RecordSchema too. The adaptation to those changes in drafts-resources are in the PR: https://github.com/inveniosoftware/invenio-drafts-resources/pull/138

PR #237, opened 2021-04-26
systemfields: add helper methods for copying files

lnielsen:


invenio_records_resources/records/systemfields/files/field.py
```diff
@@ -108,40 +108,33 @@ def pre_commit(self, record):
...
-                    # TODO: Setting these bumps the row's `revision_id`
-                    setattr(record, self._bucket_id_attr, bucket.id)
-                    setattr(record, self._bucket_attr, bucket)
-        self.store(record, FilesManager(
+        files = FilesManager(
             record,
             file_cls=self.file_cls,
             enabled=self._enabled,
-        ))
+            options=self._manger_options,
```
fenekku, line 115: minor: typo `self._manger_options` -> `self._manager_options

fenekku (approved):
Agreed, we should be mostly fine. Some things on my side might in fact be redundant, but we can remove them later.

PR #242, opened 2021-05-06
systemfields: relations caching

lnielsen:
closes inveniosoftware/invenio-app-rdm#833


invenio_records_resources/services/records/facets/facets.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def __init__(self, label=None, value_labels=None, **kwargs):
+        """Initialize class."""
+        self._label = label or ''
+        self._value_labels = value_labels
+        super().__init__(**kwargs)
+
+    def get_value(self, bucket):
+        """Get key value for a bucket."""
+        return getattr(bucket, "key_as_string", bucket.key)
```
ppanero, line 26: why `key_as_string` and not `label`? it's the key it's looking for?

invenio_records_resources/services/records/facets/facets.py
```diff
@@ -0,0 +1,246 @@
...
+        label_map = self.get_label_mapping(data.buckets)
+        for bucket in data.buckets:
+            key = self.get_value(bucket)
+            out.append({
+                "key": key,
+                "doc_count": self.get_metric(bucket),
+                "label": label_map[key],
+                "is_selected": self.is_filtered(key, filter_values)
+            })
+        return {'buckets': out, 'label': str(self._label)}
```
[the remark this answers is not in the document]
ppanero: I agree with this, though I would keep the two functions to get labels or not. Not sure if it's worth exacting code while keeping to functions (i.e. due to calls will end up in the same amount of code.

invenio_records_resources/services/records/facets/facets.py
```diff
@@ -0,0 +1,250 @@
...
+    def add_filter(self, filter_values):
+        """Construct a filter query for the facet."""
+        if not filter_values:
+            return
+
+        parsed_values = list(self._parse_values(filter_values).items())
+
+        f = self.get_value_filter(parsed_values[0])
+        for v in parsed_values[1:]:
+            f |= self.get_value_filter(v)
```
ppanero, line 197: nice with `|=`

ppanero (approved):
LGTM, but I'll need to "use" it to fully grasp some stuff

invenio_records_resources/services/records/facets/facets.py
lnielsen: This is because we parse all facets as strings. E.g. ``/api/records?is_published=1`` ends up as ``{'is_published': ['1']}`` when parsed. The string is then sent to Elasticsearch as a term query on a boolean field, and then Elasticsearch complains.

Honestly, the right solution here is likely that facets url arguments are explicitly parsed and validated for a specific type. I had to do above, in order to get the Status facet to work, and is equivalent to what we did before.

Before, we output the ``key_as_string`` in the aggregations, and RSK would use that key to query with. Now, I just moved this logic down into the service, so that we stay with having only one ``key`` field in the aggregations.

PR #255, opened 2021-06-04
facets: fix recursive call in nested term labels

### LINKED ISSUE #254: nested term facet get_value issue
Shouldn't [this be `get_value`](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/records/facets/facets.py#L223) instead of `get_labelled_values` ?

ppanero:
closes #254


fenekku (approved):
Honestly I am not terribly familiar, but it sounds right if we want to recurse.

PR #266, opened 2021-06-18
factories: enable configuration of the pid provider class

### LINKED ISSUE #265: factory: support pid provider customization on RecordTypeFactory
In the `factories/factory.py:RecordTypeFactory` the `record_cls` gets a [default (hardcoded) PID provider](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/factories/factory.py#L131). It should be customizable (i.e. a parameter when calling the factory.

ppanero:
closes #265

### DIFF
--- a/invenio_records_resources/factories/factory.py
+++ b/invenio_records_resources/factories/factory.py
@@ -18,6 +18,7 @@ from invenio_records_permissions import RecordPermissionPolicy
 
 from invenio_records_resources.records.api import Record
 from invenio_records_resources.records.systemfields import IndexField, PIDField
+from invenio_records_resources.records.systemfields.pid import PIDFieldContext
 from invenio_records_resources.resources import RecordResource, \
     RecordResourceConfig
 from invenio_records_resources.services import RecordService, \
@@ -45,7 +46,7 @@ class RecordTypeFactory(object):
     def __init__(
         self,
         record_type_name,
-        record_type_service_schema,
+        service_schema,
         schema_version="1.0.0",
         endpoint_route=None,
         record_dumper=None,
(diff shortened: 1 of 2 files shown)


invenio_records_resources/factories/factory.py
```diff
@@ -124,11 +133,18 @@ def create_metadata_model(self):
 
     def create_record_class(self):
         """Create record class."""
+        pid_field = PIDField(
```
ppanero, line 136: Is it worth creating a `PIDField` factory, and just make the full field configurable rather than passing many parameters? This time are 3, maybe later on are more... @lnielsen

invenio_records_resources/factories/factory.py
```diff
@@ -182,8 +198,7 @@ def create_service_class(self):
         if self.service_components:
             config_cls_attributes.update(
                 {
-                    "components": RecordServiceConfig.components
-                    + self.service_components
+                    "components": self.service_components
```
ppanero, line 201: If the user wants to get the parent ones s/he can inherit. Otherwise, there is no way of e.g. not get the `MetadataComponent`.

invenio_records_resources/factories/factory.py
lnielsen: I agree......i see two options. 1) override the entire pidfield or 2) have two parameters:

```
pid_provider_cls
pid_field_kwargs
```

lnielsen (commented):
Overall fine, just decide on if we override the entire pidfield or the two parameters.

invenio_records_resources/factories/factory.py
ppanero: I opted by the two args to:
1- Keep consistency on the naming (i.e. have a `_cls`)
2- Avoid possible problems on lazyness/partial because it would need an instance instead of a class created at config level.

PR #273, opened 2021-07-13
i18n: mark and extract strings for translation

mb-wali:
Todo:
* create a new resource for this module in transifex
* push the .pot file to created resource on transifex

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,7 +56,6 @@ coverage.xml
 
 # Translations
 *.mo
-*.pot
 
 # Django stuff:
 *.log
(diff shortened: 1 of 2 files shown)


invenio_records_resources/resources/errors.py
```diff
@@ -102,50 +103,52 @@ class ErrorHandlersMixin:
         QuerystringValidationError: create_error_handler(
             HTTPJSONException(
                 code=400,
-                description="Invalid querystring parameters.",
+                description=_("Invalid querystring parameters."),
```
zzacharo, line 106: Maybe we skip marking the error messages for now until we decide on the approach?

PR #282, opened 2021-08-03
files: validate init files via Schema

fenekku:
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/781

### DIFF
--- a/invenio_records_resources/services/files/components/metadata.py
+++ b/invenio_records_resources/services/files/components/metadata.py
@@ -12,6 +12,7 @@ from copy import deepcopy
 
 from invenio_files_rest.models import ObjectVersion
 
+from ..schema import InitFileSchema
 from .base import FileServiceComponent
 
 
@@ -20,8 +21,9 @@ class FileMetadataComponent(FileServiceComponent):
 
     def init_files(self, id, identity, record, data):
         """Init files handler."""
-        # TODO: Load via marshmallow schema?
-        for file_metadata in data:
+        schema = InitFileSchema(many=True)
+        validated_data = schema.load(data)
+        for file_metadata in validated_data:
             temporary_obj = deepcopy(file_metadata)
             record.files.create(temporary_obj.pop('key'), data=temporary_obj)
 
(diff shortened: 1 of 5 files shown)


invenio_records_resources/services/files/schema.py
```diff
@@ -10,12 +10,40 @@
...
+          that we will need to support.
+    Cons: Change a few tests, disable PUT endpoint really
+
+    Option 2: Allow extra fields
+    Pros: Everything stays the same.
+    Cons: The same is loose / not quite consistent.
+
+    Given LTS, going for option 2 so that many changes are not introduced.
+    But ideally option 1 seems better: we can add other fields when we do
+    support third-party data hosting (and perhaps become FileSchema).
```
ppanero, line 36: I would put this explanation in the PR, but not commit it to the code. It might stay there for years and not have meaning by then. A shorter version e.g. "Including all values since there is not a defined schema for files metadata" might be enough.

invenio_records_resources/services/files/components/metadata.py
```diff
@@ -20,8 +21,9 @@ class FileMetadataComponent(FileServiceComponent):
 
     def init_files(self, id, identity, record, data):
         """Init files handler."""
-        # TODO: Load via marshmallow schema?
-        for file_metadata in data:
+        schema = InitFileSchema(many=True)
```
ppanero, line 24: comment: ha the `many=True` tricked me. I was thinking how a list can pass the loading step if there is no `Nested`/`List` in the schema. Nice!

ppanero (approved):

invenio_records_resources/services/files/components/metadata.py
fenekku: Yeah it's a nice trick!

invenio_records_resources/services/files/schema.py
fenekku: I am experimenting here and it would be interesting to keep it. I will be bringing this up in the retrospecitve, but I have found it extremely hard to track down these explanations because they were outside the code (the line history can easily change and honestly it's faster/clearer to have them coupled with the code). Because I (or potential other developers) can't necessarily ask someone during work hours, it also helps with the time difference. 

It can definitely be shorter next time around :sweat_smile: .

PR #291, opened 2021-11-08
links: use record's public id i.e. record.id

fenekku:
This fixes some issue experienced generating links when the `record` doesn't use a `pid` system field to generate its public id. (e.g. in requests / request events)


fenekku:
Don't merge this in. I wasn't able to install the dependencies locally (too long), so I was hoping the CI would be able, but the CI is breaking on something else (I think?). Needs further investigation.

fenekku:
Faking the interface in invenio-requests is the way to go until this is better defined.

PR #292, opened 2021-11-17
service: add "unit of work" for atomic operations

lnielsen:
* Adds a new Unit of Work context manager used to allow atomic
  operations over multiple service calls.

* Keeps API as backwards compatible as possible.

### DIFF
--- a/invenio_records_resources/services/base/service.py
+++ b/invenio_records_resources/services/base/service.py
@@ -60,10 +60,17 @@ class Service:
 
     def run_components(self, action, *args, **kwargs):
         """Run components for a given action."""
-        # Run post commit components
+        uow = kwargs.pop("uow", None)
+
         for component in self.components:
             if hasattr(component, action):
+                # Done like this to avoid breaking API changes.
+                # uow should eventually be passed directly to the component
+                # so service/component method signature matches.
+                if uow is not None:
+                    component.uow = uow
                 getattr(component, action)(*args, **kwargs)
+                component.uow = None
 
     #
     # Units of transaction methods (creation...)
--- a/invenio_records_resources/services/files/service.py
+++ b/invenio_records_resources/services/files/service.py
@@ -9,10 +9,9 @@
 
 """File Service API."""
 
-from invenio_db import db
-
 from ..base import LinksTemplate, Service
 from ..records.schema import ServiceSchemaWrapper
+from ..uow import RecordCommitOp, unit_of_work
 
 
 class FileService(Service):
@@ -79,15 +78,12 @@ class FileService(Service):
             links_item_tpl=self.file_links_item_tpl(id_),
         )
 
-    def init_files(self, id_, identity, data):
+    @unit_of_work()
+    def init_files(self, id_, identity, data, uow=None):
         """Initialize the file upload for the record."""
         record = self.get_record(id_, identity, "create_files")
 
-        self.run_components("init_files", id_, identity, record, data)
-
-        db.session.commit()
-
-        self.run_components("post_init_files", id_, identity, record, data)
+        self.run_components("init_files", id_, identity, record, data, uow=uow)
 
         return self.file_result_list(
             self,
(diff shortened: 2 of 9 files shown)


invenio_records_resources/services/base/service.py
```diff
@@ -60,11 +60,21 @@ def components(self):
...
-        # Run post commit components
         for component in self.components:
             if hasattr(component, action):
                 getattr(component, action)(*args, **kwargs)
 
+    def run_components_with_uow(self, action, uow, *args, **kwargs):
+        """Run components for a given action."""
+        for component in self.components:
+            if hasattr(component, action):
+                # Done like this to avoid API changes.
```
lnielsen, line 71: Not pretty. Alternatively, we have to duplicate the ``components`` property.

invenio_records_resources/services/files/service.py
```diff
@@ -204,13 +192,10 @@ def delete_all_files(self, id_, identity):
...
-        self.run_components("delete_all_files", id_, identity, record, results)
+        self.run_components_with_uow(
+            "delete_all_files", uow, id_, identity, record, results)
 
-        record.commit()
-        db.session.commit()
-
-        self.run_components(
-            "post_delete_all_files", id_, identity, record, results)
+        uow.register(RecordCommitOp(record))
```
lnielsen, line 198: Actually, there might be a bug here because it's not indexing the record afterward.

invenio_records_resources/services/records/service.py
```diff
@@ -218,15 +218,17 @@ def reindex(self, identity, params=None, es_preference=None, **kwargs):
...
         """Create a record.
 
         :param identity: Identity of user creating the record.
         :param data: Input data according to the data schema.
         """
-        return self._create(self.record_cls, identity, data)
+        return self._create(self.record_cls, identity, data, uow=uow)
 
-    def _create(self, record_cls, identity, data, raise_errors=True):
+    @unit_of_work()
```
lnielsen, line 230: This is here, so that we're sure we always have a unit of work.....it will do nothing when called from ``create()`` above.

invenio_records_resources/services/base/service.py
ntarocco: Do you refer to the line of code `component.uow = uow`? 
Instead of creating a new `run_components_with_uow`, can't you do in `run_components`:
```
uow  = kwargs.get("ouw")
if uow:
    component.uow = uow
    ....
```

invenio_records_resources/services/uow.py
```diff
@@ -0,0 +1,246 @@
...
+        self._index_refresh = index_refresh
+
+    def on_register(self, uow):
+        """Commit record (will flush to the database)."""
+        self._record.commit()
+
+    def on_commit(self, uow):
+        """Run the operation."""
+        if self._indexer is not None:
+            arguments = {"refresh": True} if self._index_refresh else {}
```
ntarocco, line 109: Nice to have the `refresh` option here ;)

invenio_records_resources/services/uow.py
```diff
@@ -0,0 +1,246 @@
...
+
+    def __init__(self, record, indexer=None, index_refresh=False):
+        """Initialize the record commit operation."""
+        self._record = record
+        self._indexer = indexer
+        self._index_refresh = index_refresh
+
+    def on_register(self, uow):
+        """Commit record (will flush to the database)."""
+        self._record.commit()
```
ntarocco, line 104: I am not sure I understand why we `commit` on `register`, but then `index` on `commit` below?

invenio_records_resources/services/files/service.py
```diff
@@ -79,15 +78,13 @@ def list_files(self, id_, identity):
             links_item_tpl=self.file_links_item_tpl(id_),
         )
 
-    def init_files(self, id_, identity, data):
+    @unit_of_work()
```
ntarocco, line 81: I like that we call it as what it does, but I am wondering if it should be maybe more specific? E.g. `commit_after` or `persist_after` (sorry, no better naming at the moment...)
At the same time, such name reminds me of [invenio-deposit](https://github.com/inveniosoftware/invenio-deposit/blob/master/invenio_deposit/api.py#L187), and we never liked that...

PR #296, opened 2022-01-20
Lift up entity reference stuff from Invenio-Requests

max-moser:
Partially closes https://github.com/inveniosoftware/invenio-requests/issues/12

In a follow-up task, we should probably adjust the owners structure in the parent records' access field (https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/records/systemfields/access/owners.py) to use the newly created mechanisms

### DIFF
--- a/invenio_records_resources/records/systemfields/calculated.py
+++ b/invenio_records_resources/records/systemfields/calculated.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 - 2022 TU Wien.
+#
+# Invenio-Records-Resources is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Systemfield for calculated properties of records."""
+
+
+import abc
+
+from invenio_records.systemfields import SystemField
+
+
+class CalculatedField(SystemField, abc.ABC):
+    """Systemfield for returning calculated properties."""
+
+    def __init__(self, key=None, use_cache=False):
+        """Constructor."""
+        super().__init__(key)
+        self._use_cache = use_cache
+
+    def obj(self, instance):
+        """Calculate and return the record's property."""
+        # per default, no caching should be used
+        if not self._use_cache:
+            return self.calculate(instance)
+
+        # check the cache
+        obj = self._get_cache(instance)
+        if obj is not None:
+            return obj
+
+        # calculate and set cache
+        obj = self.calculate(instance)
+        self._set_cache(instance, obj)
+        return obj
+
+    def __get__(self, record, owner=None):
+        """Calculate and return the record's property."""
+        if record is None:
+            # access by class
+            return self
+
+        return self.obj(record)
+
+    def __set__(self, record, value):
+        """Prevent setting of a calculated property."""
+        msg = f"Cannot set value for calculated field '{self.key}'"
+        raise AttributeError(msg)
+
+    @abc.abstractmethod
+    def calculate(self, record):
+        """Logic for calculating the record's property."""
+        return None
--- a/invenio_records_resources/records/systemfields/entity_reference.py
+++ b/invenio_records_resources/records/systemfields/entity_reference.py
@@ -1,0 +1,123 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 - 2022 TU Wien.
+#
+# Invenio-Records-Resources is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Systemfield for managing referenced entities in request."""
+
+from functools import partial
+
+from invenio_records.systemfields import SystemField
+
+from ...references.resolvers import EntityProxy
+
+
+class ReferencedEntityField(SystemField):
+    """Systemfield for managing the request type."""
+
+    def __init__(
+        self, key=None, reference_check_func=None, resolver_registry=None
+    ):
+        """Constructor."""
+        super().__init__(key=key)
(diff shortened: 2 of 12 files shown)


fenekku:
I've created the follow-up ticket here: https://github.com/inveniosoftware/invenio-rdm-records/issues/873

invenio_records_resources/records/systemfields/entity_reference.py
```diff
@@ -0,0 +1,119 @@
...
+
+    In case the ``ref_dict`` is ``None``, it will check if this is allowed
+    for the reference at hand.
+    Otherwise, it will check if the reference dict's key (i.e. the TYPE)
+    is allowed.
+    """
+    if ref_dict is None:
+        return get_allows_none(request)
+
+    ref_type = list(ref_dict.keys())[0]
```
fenekku, line 96: If ref_dict is an empty `{}`, we'll get an index error here. (`{} is None` will give us `False`).
Maybe the check above should just be `if ref_dict`, but I am not sure I know what `check_allowed_references` is really doing in that scenario ... ?

invenio_records_resources/services/references/schema.py
```diff
@@ -0,0 +1,53 @@
...
+    """Base schema for entity references, allowing only a single key.
+
+    Example of an allowed value: ``{"user": 1}``.
+    Example of a disallowed value: ``{"user": 1, "record": "abcd-1234"}``.
+    """
+
+    class Meta:
+        """Schema meta."""
+
+        unknown = RAISE
```
fenekku, line 28: This is the default, so we don't have to specify it anymore: https://marshmallow.readthedocs.io/en/stable/quickstart.html#handling-unknown-fields  .

invenio_records_resources/references/resolvers/base.py
```diff
@@ -0,0 +1,189 @@
...
+
+    @abstractmethod
+    def _reference_entity(self, entity):
+        """The logic building a reference dict from the given entity.
+
+        Since the compatibility checks are already taken care of, this method
+        can assume that the entity is compatible and can focus on simply
+        creating the reference dict for the given entity.
+        """
+        return None
```
fenekku, line 189: comment: I think those will be much clearer to me when concrete test cases highlight what they do.

invenio_records_resources/services/references/schema.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @validates_schema
+    def there_can_be_only_one(self, data, **kwargs):
```
fenekku, line 31: :crossed_swords: :smile:

fenekku (commented):
Just the `{}` case (and eventually examples to really understand).

invenio_records_resources/records/systemfields/calculated.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def __set__(self, record, value):
+        """Prevent setting of a calculated property."""
+        msg = f"Cannot set value for calculated field '{self.key}'"
+        raise AttributeError(msg)
+
+    @abc.abstractmethod
+    def calculate(self, record):
+        """Logic for calculating the record's property."""
+        return None
```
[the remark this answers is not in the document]
max-moser: sounds good to me, a bit of renaming could make it much clearer what the intended purpose of all the methods are.
in a few cases, i think it could also make sense to merge the `__get__` and `obj` methods, to reduce the number of methods being defined...
unless of course if the `obj` method is part of the general interface, though i couldn't see any place where that would be relied on

invenio_records_resources/records/systemfields/entity_reference.py
max-moser: good point; even though it shouldn't (as this is checked in a couple places, e.g. the marshmallow schema) be the case that a ref dict is something else other than `None` or a dictionary with exactly one key/value pair (the "type" as the key, the ID as value; e.g. `{"user": "1"}`), an additional check here doesn't hurt and at the very least makes the expected shape more clear to devs.

invenio_records_resources/references/resolvers/base.py
max-moser: basically, the Resolver Registry is a translation layer between actual entities (e.g. User or Record objects) and reference dicts (e.g. `{"user": "1"}`, or `{"record": "abcd-1234"}`) the where you can dynamically register the handlers tasked with performing the actual translation (i.e. the Entity Resolvers)... it's similar to a dynamic factory pattern, i suppose.

the Entity Resolver interface is there to enable dumping and resolving of all kinds of entities with different structures.
for instance, User entities have their ID stored in an `.id` field, but for records it might be more appropriate to use the `recid` rather than their UUID (`.id`) for reference dicts.
so in this case, you can write a custom Entity Resolver and register it to the Resolver Registry, to enable dumping and resolving of these kinds of entities.

and the Entity Proxies are buffers in the "read"/"resolve" direction (ref-dict to entity), to postpone potentially heavy DB lookup operations until you *really* want to handle the referenced entity (`proxy_obj.resolve()`), mostly for performance reasons (cf. [owners in rdm-records](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/records/systemfields/access/owners.py#L43))

invenio_records_resources/records/systemfields/calculated.py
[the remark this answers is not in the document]
fenekku: I will let you make the ticket.

invenio_records_resources/references/resolvers/base.py
fenekku: Make more sense now, thanks.

PR #300, opened 2022-02-03
[POC]: message bus implementation

ppanero:


invenio_records_resources/services/uow.py
```diff
@@ -215,6 +219,10 @@ def __init__(self, session=None):
         """Initialize unit of work context."""
         self._session = session or db.session
         self._operations = []
+        self._event_bus = EventBus(
+            handlers=current_app.config["RECORDS_RESOURCES_EVENT_HANDLERS"],
```
ppanero, line 223: should be the event bus to load this.

invenio_records_resources/services/events/bus.py
```diff
@@ -0,0 +1,34 @@
...
+
+class EventBus:
+    """Event bus."""
+
+    def __init__(self, handlers, queue):
+        """Constructor."""
+        self._handlers = handlers
+        self._queue = queue
+
+    def publish(self, event):
```
ppanero, line 22: publish should store locally events (memory), to accomodate cases of the uow failing (i.e. rollback should not keep events on the queue).

It can implement a `flush` method that will dump to the actual queue.

tests/services/events/conftest.py
```diff
@@ -0,0 +1,37 @@
...
+        queue=queue_name,
+        type="direct",
+        delivery_mode="persistent",  # in-memory and disk
+    )
+
+    app_config["QUEUES_DEFINITIONS"] = [
+        {"name": queue_name, "exchange": exchange}
+    ]
+
+    app_config["CELERY_ACCEPT_CONTENT"] = ["json", "msgpack", "yaml", "pickle"]
```
ppanero, line 35: @slint the only way I found was to globally accept pickle.... which might be a problem...

tests/services/events/test_bus.py
```diff
@@ -0,0 +1,25 @@
...
+from invenio_records_resources.services.events import EventBus
+from invenio_records_resources.services.events.events import RecordCreatedEvent
+
+
+def test_bus_publish_consume(app):
+    bus = EventBus("test-events")
+    event = RecordCreatedEvent(created=datetime.now(), recid="12345-abcde")
+
+    bus.publish(event)
+    sleep(10)
```
ppanero, line 23: Sometimes fails, sometimes doesn't I think we are falling once again in the issues we had with ES (fixed now with .index.refresh()).

invenio_records_resources/services/events/handlers.py
```diff
@@ -0,0 +1,81 @@
...
+    """Handle events queue.
+
+    :param max_events: maximum number of events to process by the task.
+    :param ttl: time to live (in seconds) for the task.
+    """
+    bus = EventBus(queue_name)
+    start = datetime.timestamp(datetime.now())
+    end = start
+    spawn_new = False
+    with bus.active_consumer() as consumer:
```
ppanero, line 70: The idea here is to have one open consumer for the duration of the task. We could potentially `consume()` but it might become 1000 consumers, even if they are light objects I think it is safe to open an active one here.

invenio_records_resources/services/events/handlers.py
```diff
@@ -0,0 +1,81 @@
...
+        while max_events > 0 and (start + ttl) > end:
```
ppanero, line 71: I did not find any configuration that would allow to kill a consumer after a determined amount of events or time elaspsed (in a similar fashion thatn uWSGI does). So I have implemented it a bit roughly for PoC purposes.

invenio_records_resources/services/events/handlers.py
```diff
@@ -0,0 +1,81 @@
...
+            spawn_new = False
+            event = consumer.consume()  # blocking
+            _handle_event(event)  # execute all handlers
+            end = datetime.timestamp(datetime.now())
+            spawn_new = True
+
+    if spawn_new:
```
ppanero, line 78: Meaning, 
- If the worker died while consuming events spawn a new task, since there are events to be consumed, and is better not to wait.
- Otherwise, it died while waiting for events, in which case is better to wait until celery beat spawns it again.

tests/services/events/conftest.py
```diff
@@ -0,0 +1,58 @@
...
+
+from invenio_records_resources.services.events.events import RecordEvent, \
+    RecordCreatedEvent
+
+
+@pytest.fixture(scope="module")
+def app_config(app_config):
+    """Application configuration."""
+    # handlers
+    app_config["RECORDS_RESOURCES_EVENTS_HANDLERS"] = {
```
ppanero, line 27: Most of this configuration would go to `invenio-events` and be overwritten in `invenio-app-rdm`. It is here for demo purposes.

max-moser:
# Summary of a brief recent discussion with Pablo about the event bus topic

The topic of the event bus is a reoccurring one.
Every now and then a new feature comes along that would benefit from a sort of event bus.
Right now, the new feature in question would be `Invenio-Stats`.


There already exist [minimal implementations](https://github.com/inveniosoftware/rfcs/pull/62/files) of event-bus-like constructs that
could potenitally be extended into a full-fledged event bus.
As such, coming up with a basic construct for an event bus shouldn't be too much of an issue - getting it right however *would be*.


## Comparison with other concepts

### Service components

Service components are responsible for performing certain actions during service methods (e.g. populating some metadata fields in records, 
resolving referenced entities, etc.).
Probably some, but not all, hooks (like *after publish*) could be replaced by the event bus.


### Units of work (UoW)

Units of work serve a different purpose, namely grouping together sets of actions into atomic units of work that either succeed of fail as a whole.
If anything, they could be used for registering an event emission operation *on commit*, as one way to generate events.


## Downsides

### Event-driven confusion

Invenio would become more event-driven, which it wasn't really before.
This could cause confusion among developers because it is not immediately apparent which pieces of code are executed during certain operations.

One way to counteract this downside would be to add a CLI command that parses the registered event handlers and prints them in a human-readable form, e.g.:
```
$ invenio events show-handlers
EVENT_NAME_1:
  Handler (async): invenio_event_bus.handlers.handle_event_1
  Handler (sync):  invenio_event_bus.handlers.handle_event_1_more

EVENT_NAME_2:
  ...
```
Note that it's easy to imagine funky ways of registering event handlers that wouldn't be picked up by this CLI command (e.g. the now deprecated `@before_first_request`).
However, if somebody writes (or uses) an Invenio extension that does something like that, we can assume that they know what they're doing anyway.


### Ensuring that no events/handlers are forgotten or executed multiple times

One of the major difficulties here would be to make sure that no events are forgotten and/or that none of the registered handlers are executed less or
more often than they should be.
Celery, RabbitMQ and the other tools in use could already provide some help here, but I'm not familiar enough with them yet to say for sure.

If the tools don't provide the necessary means, I had the idea of a "journal" for the status of events/handlers that could be stored e.g. in the cache (Redis).
This could help us keep track of which event handlers have already finished their jobs and which would need to re-run in the event of a system crash.
Example entry:
```
{"event": NAME, "handler_status": {"handler_name_1": "started", "handler_name_2": "completed"}}
```
This is only a rough idea though and still needs more thought (e.g. what if the cache dies too).



## Alternatives (for Invenio-Stats)

We could create yet another sort of minimal implementation (similar to the update propagation) where we add a UoW operation that
directly acts as stats event emitter on commit.


# Further remarks

It's already evening and I should go home.
Here are some further remarks that I found again too late to really rework into this summary:

question: do we support multiple ways of running, or we assume handlers are run sequentially for an event?  
IIRC, the current implementation might have an issue. It launches new celery tasks, which means that order might not be respected and we might have dirty reads.  
However, tackling this kind of problems seems a bit overkill. What about documenting them, then choosing a compromise and adhering to it. For this, we need a list of use cases indeed.

# Conclusion

The event bus would be an awesome feature to have, but needs some more investigation and proper design, as well as development time for getting it _right_.

I think we should create a list of "relevant dangerous scenarios" that we want to safeguard against, and come up with ideas of how to prevent them, or
recover from them.

This could also provide an opportunity to have a look at the state of (celery) tasks in Invenio in general.
AFAIK failures in (scheduled) tasks aren't really monitored or logged, but rather follow a kind of fire-and-forget philosophy.

invenio_records_resources/services/events/bus.py
```diff
@@ -0,0 +1,43 @@
...
+        self._queue = None
+
+        for name, queue in current_queues.queues.items():
+            if name == self._queue_name:
+                self._queue = queue
+                break
+
+    def publish(self, event):
+        """Publish an event to the bus queue."""
+        return self._queue.publish([dumps(event)])
```
ntarocco, line 33: I would make configurable the serialization obj and not hardcode `pickle`. Celery does the same. We might want to use simple JSON or `msgpack` instead.

invenio_records_resources/services/events/events.py
```diff
@@ -0,0 +1,40 @@
...
+    recid: str
+    type: ClassVar[str] = "RECORD"
+    handling_key: ClassVar[str] = "RECORD"
+
+
+@dataclass
+class RecordCreatedEvent(RecordEvent):
+    """Record related events."""
+
+    action: ClassVar[str] = "PUBLISHED"
```
ntarocco, line 39: if `RecordCreatedEvent` is always for published, not sure that the `action` field is useful?

Is the `handling_key` necessary? Isn't it the class name/import enough? Unless there could be multiple class events with the same `handling_key`.

invenio_records_resources/services/events/handlers.py
```diff
@@ -0,0 +1,81 @@
...
+
+    for handler in handlers:
+        func = handler
+        async_ = True
+        if isinstance(handler, tuple):
+            func = handler[0]
+            async_ = handler[1]
+
+        if async_:
+            func.delay(**asdict(event))
```
ntarocco, line 49: We may want to pass more params to the `delay` func: wouldn't it be better to always have a sync func to handle the event, and that it eventually spawns an async task?
```
def hander(...):
    my_async_handler.delay(....)
```

invenio_records_resources/services/events/handlers.py
```diff
@@ -0,0 +1,81 @@
...
+        else:
+            func(**asdict(event))
+
+        # audit logging
```
ntarocco, line 53: It is not really audit logging this one. I would suggest creating a dedicate logger for the events so it can be configured ad-hoc.

invenio_records_resources/services/events/bus.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def __init__(self, queue_name=None):
+        """Constructor."""
+        self._queue_name = queue_name or \
+            current_app.config["RECORDS_RESOURCES_EVENTS_QUEUE"]
+        self._queue = None
+
+        for name, queue in current_queues.queues.items():
+            if name == self._queue_name:
+                self._queue = queue
```
ntarocco, line 28: Will the queue always be found, or can it exit the for loop and stay `None`?

invenio_records_resources/services/events/handlers.py
ntarocco: is there a specific need for this?

ntarocco (commented):
Good job! Added my 2 cents here.
I would also move all this to `flask-resources`, if not to a specific module.

PR #308, opened 2022-03-11
factory: customize metadata class attributes

ppanero:
closes https://github.com/inveniosoftware/invenio-vocabularies/issues/156

## Changes:

**Factory**
- Support customization of model class attributes
- Support customization of record class attributes
- Support customization of pid field constructor parameters

**Systemfields**
- Add `ModelPIDField`: lightweight solution to avoid overloading PIDStore. This field is also in charge of managing its status, stored in another column (e.g. `pid_status`)

### DIFF
--- a/invenio_records_resources/factories/factory.py
+++ b/invenio_records_resources/factories/factory.py
@@ -58,7 +58,8 @@ class RecordTypeFactory(object):
         permission_policy_cls=None,
         pid_field_cls=PIDField,
         pid_field_kwargs=None,
-        model_cls_attrs=None
+        model_cls_attrs=None,
+        record_cls_attrs=None,
     ):
         """Constructor."""
         self.record_type_name = record_type_name
@@ -81,6 +82,7 @@ class RecordTypeFactory(object):
         self.schema_path = self._build_schema_path(schema_path)
         self.index_name = self._build_index_name(index_name)
         self.model_cls_attrs = model_cls_attrs or {}
+        self.extra_record_cls_attributes = record_cls_attrs or {}
 
         # resource class attributes
         self.endpoint_route = endpoint_route
@@ -152,6 +154,8 @@ class RecordTypeFactory(object):
             "pid": pid_field,
             "dumper": self.record_dumper or ElasticsearchDumper(),
         }
+
+        record_class_attributes.update(self.extra_record_cls_attributes)
 
         if self.record_relations:
             record_class_attributes["relations"] = self.record_relations
--- a/invenio_records_resources/factories/factory.py
+++ b/invenio_records_resources/factories/factory.py
@@ -68,6 +68,7 @@ class RecordTypeFactory(object):
         # pid field attributes
         self.pid_field_cls = pid_field_cls
         self.pid_field_kwargs = pid_field_kwargs or {
+            "key": "id",
             "create": True,
             "provider": RecordIdProviderV2,
             "context_cls": PIDFieldContext,
@@ -142,8 +143,7 @@ class RecordTypeFactory(object):
 
     def create_record_class(self):
         """Create record class."""
-        pid_field = self.pid_field_cls(
-            "id", **self.pid_field_kwargs)
+        pid_field = self.pid_field_cls(**self.pid_field_kwargs)
 
         record_class_attributes = {
             "model_cls": self.model_cls,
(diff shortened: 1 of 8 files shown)


invenio_records_resources/records/systemfields/pid.py
```diff
@@ -180,3 +182,69 @@ def dump_obj(field, record, pid):
...
+        # Class access
+        if record is None:
+            return self._context_cls(self, owner)
+        # Instance access
+        try:
+            return getattr(record.model, self.model_field_name)
+        except AttributeError:
+            return None
+
+    def pre_create(self, record):
```
ppanero, line 244: I do not understand why we need this. The `funder` record has a pid attr in the root, e.g.
```json
{
    "pid": ...
    "name": ...
}
```

Then when the record gets created, in invenio-records [here](https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py#L73) the dict should get the attribute (since there is a `pid` field defined in the model). Which is how the other model fields work...

However, without this `pre_create` the tests fail because of a `pid` constraint null (i.e. the pid was not send to the DB in the INSERT query).

invenio_records_resources/records/systemfields/pid.py
```diff
@@ -180,3 +183,69 @@ def dump_obj(field, record, pid):
...
+        """Resolve identifier."""
+        resolver = self.field._resolver_cls(
+            self._record_cls, self.field.model_field_name
+        )
+        pid, record = resolver.resolve(pid_value)
+        self.field._set_cache(record, pid)
+
+        return record
+
+    def create(self, record):
```
ppanero, line 210: Needed so the `PIDComponent` can create it. I am slightly puzzled this is needed. In my opinion, the `pid` attribute should be set on `invenio-records` when the "dict" is initialize (`RecordBase` constructor), removing the need for a PIDComponent in the first place. However, it does not seem to be being set (`ModelField` should do this right?)

invenio_records_resources/factories/factory.py
```diff
@@ -67,6 +69,7 @@ def __init__(
         # pid field attributes
         self.pid_field_cls = pid_field_cls
         self.pid_field_kwargs = pid_field_kwargs or {
+            "key": "id",
```
ppanero, line 72: remove this

ppanero, line 72: remove this

invenio_records_resources/factories/factory.py
```diff
@@ -78,6 +81,8 @@ def __init__(
         self.record_relations = record_relations
         self.schema_path = self._build_schema_path(schema_path)
         self.index_name = self._build_index_name(index_name)
+        self.model_cls_attrs = model_cls_attrs or {}
+        self.extra_record_cls_attributes = record_cls_attrs or {}
```
slint, line 85:
```suggestion
        self.record_cls_attrs = record_cls_attrs or {}
```

I thought the different naming was because of a name-clash, but it seems that there's `self.record_cls_attrs`

invenio_records_resources/records/systemfields/pid.py
```diff
@@ -180,3 +184,83 @@ def dump_obj(field, record, pid):
...
+        """Method to create a new persistent identifier for the record."""
+        # pop from metadata
+        pid_value = record.pop(self.field.model_field_name)
+        self.field._provider.create(
+            pid_value, record, self.field.model_field_name
+        )
+        # setattr(record, self.field.model_field_name, pid)
+        # record.pid_status = PIDStatus.REGISTERED  # trigger __set__
```
slint, line 219: I guess these can be removed (I see they're handled in the provider)

slint (commented):
Some renames and leftover commented lines, but otherwise LGTM

invenio_records_resources/factories/factory.py
ppanero: Fixed, I just wanted to avoid having `self.record_cls_attrs` and `record_class_attributes` (which exists below). And named it extra because they are actually added (not overwritting). But name changed, is all almost the same in the end, and your proposal is consistent with the other names.

invenio_records_resources/records/systemfields/pid.py
ppanero: Yup, sorry. Leftovers...

PR #316, opened 2022-04-04
service: pass uuids to bulk indexer

ppanero:
- The `scan()` method returns elasticsearch `Hit` instances. These have the _recid_ in the `id` property. Therefore the current `res.id` would pass them to the indexer.
- The indexer expects UUIDs, to avoid making a query to pidstore (i.e. avoid using `resolve`, and use `get_record` instead).

In addition, this PR uses source filtering to lighten the load on the query, i.e. get back only the `meta.id` since it is the only value that's useful.

### DIFF
--- a/invenio_records_resources/services/records/service.py
+++ b/invenio_records_resources/services/records/service.py
@@ -211,12 +211,21 @@ class RecordService(Service):
         """Reindex records matching the query parameters."""
         self.require_permission(identity, 'search')
 
-        # Prepare and execute the search as scan()
+        # prepare and update query
         params = params or {}
-        search_result = self._search(
-            'reindex', identity, params, es_preference, **kwargs).scan()
+        params.update(kwargs)
 
-        iterable_ids = (res.id for res in search_result)
+        # create an Elasticsearch DSL, we do not want components to run
+        search = self.search_request(
+            identity,
+            params,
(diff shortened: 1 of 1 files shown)


invenio_records_resources/services/records/service.py
```diff
@@ -211,12 +211,21 @@ def reindex(self, identity, params=None, es_preference=None, **kwargs):
         """Reindex records matching the query parameters."""
         self.require_permission(identity, 'search')
 
-        # Prepare and execute the search as scan()
+        # prepare and update query
         params = params or {}
-        search_result = self._search(
-            'reindex', identity, params, es_preference, **kwargs).scan()
+        params.update(kwargs)
```
slint, line 216: Not sure about moving `kwargs` here since it was the only way before to pass things to `self.search_request(...)`. I'm aware the API is a bit "vague" and has a lot of indirection, but in that case, I would take a step back and review the overall usage before changing "core" parts.

slint: Ok, just noticed that it used to be `self._search(...)` before (vs. `self.search_request(...)`). I think as we discussed IRL it's better to go directly for the "low-level" API :)

slint (approved):

PR #318, opened 2022-04-06
registry: register indexer from service

ppanero:
- requires https://github.com/inveniosoftware/invenio-indexer/pull/133
- closes https://github.com/inveniosoftware/invenio-indexer/issues/132

### DIFF
--- a/invenio_records_resources/registry.py
+++ b/invenio_records_resources/registry.py
@@ -8,6 +8,8 @@
 
 """Service registry."""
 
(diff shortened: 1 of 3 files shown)


ppanero:
After discussions IRL, due to the fact that and app ctx might not be available at time of registration, we cannot access the indexer registry. Same would happen accessing the ext directly, since the loading order is not guaranteed. Therefore, we are moving the registration to rdm-records and separating it.

PR #323, opened 2022-04-11
global: add support for back relations update

### LINKED ISSUE #322: add support for relations update propagation
As per RFC https://codimd.web.cern.ch/96X3AjEhRLWfIyxOyWPr9g?edit

ppanero:
- requires https://github.com/inveniosoftware/invenio-records/pull/282
- requires https://github.com/inveniosoftware/invenio-indexer/pull/135
- closes #322 

Tests passing locally with indexer installed from PR:
Screenshot 2022-04-26 at 14 19 13


invenio_records_resources/services/records/components.py
```diff
@@ -133,3 +135,20 @@ class RelationsComponent(ServiceComponent):
...
+
+
+class BackRelationsComponent(ServiceComponent):
+    """Back Relations service component."""
+
+    def update(self, identity, data=None, record=None, uow=None, **kwargs):
+        """Register a task for the update propagation."""
+        # FIXME: we need a better way... service.name would be ok
+        # or registry.service_id(self.service)
+        record_type = str(record.__class__.__name__).lower() + "s"
```
ppanero, line 147: Look into this

invenio_records_resources/services/records/service.py
```diff
@@ -415,3 +415,39 @@ def rebuild_index(self, identity, uow=None):
...
+        fieldpaths = ["metadata.inner_record"]
+
+        clauses = []
+        for field in fieldpaths:
+            for record in records:
+                recid = record["id"]
+                revision_id = record["revision_id"]
+                clauses.append(Q("bool", must=[
+                    Q("term", **{f"{field}.id": recid}),
+                    # Q("term", **{f"{field}.@v": f"!='{recid}:{revision_id}'"}) # noqa
```
ppanero, line 435: missing half a query, but it should work adding that anyway

tests/services/test_service_relation_propagation.py
```diff
@@ -0,0 +1,120 @@
...
+    updated_data = deepcopy(input_data)
+    updated_data["metadata"]["title"] = "new title"
+    item = service.update(identity_simple, id_, updated_data)
+    service.record_cls.index.refresh()
+
+    read_item = service.read(identity_simple, id_)
+    assert read_item["metadata"]["title"] == "new title"
+
+    # process the index queue
+    service_wrel.indexer.process_bulk_queue()
```
ppanero, line 104: in real life this would be triggered by a celery beat task.

invenio_records_resources/services/records/components.py
lnielsen: 👍 Or simply an attribute on the record class itself.

invenio_records_resources/services/records/components.py
```diff
@@ -133,3 +135,20 @@ class RelationsComponent(ServiceComponent):
...
+        # FIXME: until the run_components has been fixed the uow
+        # is passed as a cmp attr instead of param.
+        self.uow.register(TaskOp(
```
lnielsen, line 150: So records that wants to use this change propagation, will simply add the this component in their config?

invenio_records_resources/services/records/service.py
```diff
@@ -415,3 +415,39 @@ def rebuild_index(self, identity, uow=None):
...
+                ]))
+
+        filter = [Q("range", indexed_at={"lte": notif_time})]
+        es_query = Q(
+            "bool", minimum_should_match=1, should=clauses, filter=filter
+        )
+
+        search = self.create_search(
```
lnielsen, line 443: Assuming the reindex method doesn't work properly? if so we can simply also change it to what you need it to do. I don't think it's really being used anywhere

invenio_records_resources/tasks.py
```diff
@@ -1,21 +1,42 @@
...
     service = current_service_registry.get(service_id)
     service.extract_file_metadata(system_identity, record_id, file_key)
+
+
+@shared_task(ignore_result=True)
+def propagate_relation_updates(record_type, record_ids):
+    """Execute the handlers set up for a record_type update."""
+    task_start = arrow.utcnow().strftime("%Y-%m-%dT%H:%M:%S.%f")
+
+    service = current_service_registry.get(record_type)
```
lnielsen, line 31: I think we might be able to avoid this read. 

1. Update runs task async by sending a list of user ids

The task will:
2. Read all changed users
3. Call each registered handler with the list of users.
4 A handler will use only the ids and revision ids of the users to search for related records that needs update.
5 The handler will then send each of the related record ids for bulk indexing.

...

6. Bulk indexing is happening at a later time and running it will read the user again.

Hence, if we look at a "change notification" instead of a list of ids, but list of (id,revision_id) pairs then we can avoid reading the user twice.

The trade-off is then when the propagation task is running, we potentially don't have the latest revision_id because more updates may have happened in-between step 1 and 2. However, since, since step 4 also takes the "indexed_at" time into account I think we'll on average have less select queries on the database.

lnielsen (commented):
Overall, my main comment is about when to read the changed object, and the other is naming :-)

Instead of back relations, I'd use something like "change_notificaiton".

Later PRs, we'll have to also look at deletes, but perhaps let's avoid that can of worms in the first round :-)

invenio_records_resources/services/records/components.py
ppanero: This variable value needs to match the "service name" given upon registration, currently done at views.py level in invenio-rdm-records. This leads to:
- `record_type` might not be the best variable name
- we have the need to set it somewhere, and it will be used for "registering change notification handlers", "services" and "indexers"

WDYT?

invenio_records_resources/services/records/components.py
ppanero: yes, along with the `indexed_at` (dumper ext + cmp)

invenio_records_resources/services/records/service.py
ppanero: I will look into it, it was mainly to be able to build the query with ES DSL instead of querystring which is what the `params` parameter of `reindex` accepts.

invenio_records_resources/tasks.py
ppanero: While I agree with this, the problem of the revision_id is bigger. When updating the task will be registered by the component (before the `RecordCommitOp`). Therefore the `revision_id` that would be passed would be outdated from the beginning.

- We can play with "`+1`" but is a hardcoded value that can be changed for some reason. It very easily leads to inconsistent data.
- It is true that having the `indexed_at`, we can simply "remove" the `@v` from the query. Worst case we would reindex records that got indexed in between the record.commit() and the task start (when it calculates the `arrow.utcnow`).

invenio_records_resources/services/records/components.py
slint: I like the fact that the service has become "the one place" that brings a lot of things together (record class, indexer, permissions, etc.), so IMO the service ID (i.e. `names`, `users`, etc.) feels like a good lookup key. Also we're going with components and service methods to implement these notifications/update-propagation, so we're anyways on the service level already.

I think adding `.name` or `.id` to the service, would be the best, and would also shorten the service registration call to just `current_service_registry.register(users_service)`.

invenio_records_resources/services/records/service.py
slint: Another benefit of re-using `reindex` is e.g. that it has now been optimized with `.source(False)` to return only record (UU)IDs... Maybe we also need to move it under `_reindex` (similar to `_create`) so that it can be reused more easily.

invenio_records_resources/tasks.py
lnielsen: You can probably make it work by making new uow operation:

You would pass the list of records instead of a list of id/revision_id pairs:

```python
self.uow.register(ChangeNotificationOp(record_type, [record]))
```

The task itself would the on post_commit send the celery tasks with a list of id/revision pairs (i.e. after the record has been committed and revision id bumped):

```python
class ChangeNotificationOp(Operation):
    def __init__(self, record_type, records):
        self._record_type = record_type
        self._records = records

    def on_post_commit(self, uow):
        send_change_notifications.delay(
             record_type,
             [(r.id, r.revision_id) for r in records]
        )
```

invenio_records_resources/services/records/service.py
lnielsen: We can adapt reindex to take a DSL query instead or in addition to a query string. It hasn't really been used before so ok to refactor it to make it useful. It's fine to make things work first, and later move around

invenio_records_resources/tasks.py
ppanero: I get the feeling of doing a very ad-hoc implementation. While I think this solves the problem of the `revision_id`, it can be that other notification handlers need some other data of the record. Then the question is: "is there such a use case?" if yes, then we might have to refactor this again, if not then I am not sure about generalizing to a "notification change system" instead of "relations update".

I'm just putting out there my thoughts, but I will implement it this way so we can move ahead :)

invenio_records_resources/services/records/service.py
ppanero: done, made a `es_query` para optional

invenio_records_resources/services/records/components.py
ppanero: Done, there was already a `self.config.service_id` that was allowed in the service registry. I made it consistent as part of the base config (and set it for files and records). Means we will have to add this across services but is not heavy.

lnielsen (approved):

PR #324, opened 2022-04-12
relations: added PID nested relation

alejandromumo:
This PR is needed to solve https://github.com/inveniosoftware/invenio-records-resources/issues/325


ppanero:
discussed and tested in real life, the `PIDListRelation` is enough

PR #329, opened 2022-04-27
service: set indexer routing key at producer level

ppanero:
The routing key was defined at queue level, however the publish method was using that of the producer. This lead to messages being sent to two queues. For example on community members update it was sending to `members` (queue level routing key) and `records` (producer level routing key - default)

### DIFF
--- a/invenio_records_resources/services/records/service.py
+++ b/invenio_records_resources/services/records/service.py
@@ -34,13 +34,20 @@ class RecordService(Service):
     def indexer(self):
         """Factory for creating an indexer instance."""
         return self.config.indexer_cls(
+            # the routing key is mandatory in the indexer constructor since
+            # it is afterwards passed explicitly to the created consumers
+            # and producers. this means that it is not strictly necessary to
+            # pass it to the queue constructor. however, it is passed for
+            # consistency (in case the queue is used by itself) and to help
+            # entity declaration on publish.
             queue=LocalProxy(
                 lambda: Queue(
                     self.config.indexer_queue_name,
                     exchange=current_app.config["INDEXER_MQ_EXCHANGE"],
-                    routing_key=current_app.config["INDEXER_MQ_ROUTING_KEY"],
+                    routing_key=self.config.indexer_queue_name,
                 )
             ),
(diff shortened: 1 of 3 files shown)


invenio_records_resources/services/records/service.py
```diff
@@ -34,13 +34,20 @@ class RecordService(Service):
     def indexer(self):
         """Factory for creating an indexer instance."""
         return self.config.indexer_cls(
+            # the routing key is mandatory in the indexer constructor since
```
ppanero, line 37: important explanation of why the duplicated parameter passing (queue name as routing key)

invenio_records_resources/services/records/service.py
```diff
@@ -34,13 +34,20 @@ class RecordService(Service):
...
+            # and producers. this means that it is not strictly necessary to
+            # pass it to the queue constructor. however, it is passed for
+            # consistency (in case the queue is used by itself) and to help
+            # entity declaration on publish.
             queue=LocalProxy(
                 lambda: Queue(
                     self.config.indexer_queue_name,
                     exchange=current_app.config["INDEXER_MQ_EXCHANGE"],
-                    routing_key=current_app.config["INDEXER_MQ_ROUTING_KEY"],
+                    routing_key=self.config.indexer_queue_name,
```
slint, line 47: Comment: from a reductionist POV, I would say "why not completely remove the argument", but I see that even though we just use direct queues/exchanges, we still need the routing key to be specified... I thought maybe having a default like "*" applied to the routing key could make it work always, but looks like that's too many places to change (and test that it works without side effects)

slint (approved):

invenio_records_resources/services/records/service.py
ppanero: I agree on this. As I mentioned, it could be just set at indexer level and then "trust" the queue routing key will never be used (we cannot really guarantee this). Another option would be to remove the explicit passing of the routing key to the consumer/producer. But this is a breaking change at invenio-indexer/api level that could lead to loss of tasks and inconsistent data.

Therefore, having an extra param to make sure things are consistent seems like the lesser evil 😢 😅

PR #336, opened 2022-05-10
[EXPAND] service: allows record fields expansion

ntarocco:


invenio_records_resources/resources/records/config.py
```diff
@@ -33,8 +33,8 @@ class RecordResourceConfig(ResourceConfig):
     request_view_args = {"pid_value": ma.fields.Str()}
     request_search_args = SearchRequestArgsSchema
     request_extra_args = {
-        "expand": ma.fields.Boolean(load_default=False),
-        "refresh": ma.fields.Boolean(load_default=False),
+        "expand": ma.fields.Boolean(),
```
ntarocco, line 36: This is to avoid to always have in the generated links:
`?expand=False`

Do we prefer to always have or never have it?
I went for the second option, maybe the first one is more correct, but much more ugly.

lnielsen: I agree we should not add the parameter. I think best is that links don't contain parameters, so it's up to the caller to add the parameters they like......exception is prev/next/self links for search results where we need the parameters to avoid making it difficult for the frontend to create them.

PR #337, opened 2022-05-11
[DO NOT MERGE] systemfield: make index aware of prefix

ppanero:
- closes https://github.com/inveniosoftware/invenio-communities/issues/620

### DIFF
--- a/invenio_records_resources/records/systemfields/index.py
+++ b/invenio_records_resources/records/systemfields/index.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2020 CERN.
+# Copyright (C) 2020-2022 CERN.
 # Copyright (C) 2020 Northwestern University.
 #
 # Invenio-Records-Resources is free software; you can redistribute it and/or
@@ -12,7 +12,8 @@
 from elasticsearch_dsl import Index
 from invenio_records.systemfields import SystemField
 from invenio_search import current_search_client
-
+from invenio_search.utils import build_index_name
+from werkzeug.local import LocalProxy
 
 class IndexField(SystemField):
     """Index field."""
@@ -26,7 +27,10 @@ class IndexField(SystemField):
         if isinstance(index_or_alias, Index):
             self._index = index_or_alias
         else:
-            self._index = Index(index_or_alias, using=current_search_client)
+            self._index = Index(
+                LocalProxy(lambda: build_index_name(index_or_alias)),
+                using=current_search_client
+            )
         # Set search alias name directly on the index
         self._index.search_alias = search_alias or self._index._name
 
@@ -35,4 +39,5 @@ class IndexField(SystemField):
     #
     def __get__(self, record, owner=None):
         """Get the persistent identifier."""
+
         return self._index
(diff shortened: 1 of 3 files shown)


invenio_records_resources/records/systemfields/index.py
```diff
@@ -29,10 +30,16 @@ def __init__(self, index_or_alias, search_alias=None):
...
         # Set search alias name directly on the index
         self._index.search_alias = search_alias or self._index._name
+        self._prefix_set = False
 
     #
     # Data descriptor methods (i.e. attribute access)
     #
     def __get__(self, record, owner=None):
         """Get the persistent identifier."""
+        if not self._prefix_set:
```
ppanero, line 40: This is not ideal, but we cannot access `current_app` in the constructor since it's instantiated before an application context is pushed. Aternatively, I could make the whole `self._index` a `LocalProxy`... Both feel hackish.

ppanero:
found an error, some indices were prefixed several times.
Screenshot 2022-05-11 at 14 56 59

ppanero:
This solution is not suitable. The `Index` would have the prefix and would work by itself. However, invenio-indexer API calls `_prepare_index` before performing operations (i.e. `IndexRecordOp --> indexer.index(record)`). This call will mean ending up with the prefix added twice (once from `IndexField` another from `Indexer`).

Another alternative is to add a line such as the following every time we want to get the actual index name (without being an indexer call, i.e. `IndexRefreshOp`).

```
index=self.indexer._prepare_index(self.record_cls.index)
```

TL;DR solutions:

a) Make `IndexField` the sole point of prefix/suffix expansion. This requires lots of changes and is back incompatible for invenio-indexer.
b) Be aware of using `_prepare_record` when getting the index name from the field. This requires the least changes, but we might end up in trouble in the future if we pass an already _prepared_ index to the function (prefixed twice)

My take, due to the need of a fast fix: go for `b)` adding an `if` to check if it has already been prefixed. For now, the only extra place is the `IndexRefreshOp`.

cc @lnielsen @ntarocco (can discuss tomo after standup)

ppanero:
This last fix is working (see instance indices), also tested creating a community (issue reported) and it works

Screenshot 2022-05-12 at 10 44 59

However,
1- This means that tests will never work with prefixes, since we access the `APIClass.index.refresh()` directly
2- Need to be aware if we ever do something directly on the index

Makes me think if the refresh should be delegated to `invenio-indexer`, then the `IndexRefreshOp` gets and indexer as parameter. Avoiding accessing the API classes/indices directly.

invenio_records_resources/services/uow.py
```diff
@@ -189,7 +190,7 @@ class IndexRefreshOp(Operation):
 
     def __init__(self, index, **kwargs):
         """Initialize the index to be refreshed."""
-        self._index = index
+        self._index = index.clone(name=build_alias_name(index._name))
```
ppanero, line 193: I'm slightly concerned about the performance of cloning. However,
- tests did not have an increase of time (4m50s while the previous run was 4m51s)
- the func description says _Create a copy of the instance with another name or connection alias. Useful for creating multiple indices with shared configuration_. So I'd venture to say it's for this purpose and things like the underlying connection is not duplicated.

ppanero:
Closing this PR in favour of the ones in indexer/records-resources

PR #339, opened 2022-05-12
uow: use indexer in refresh op

ntarocco:
- requires https://github.com/inveniosoftware/invenio-indexer/pull/137
- closes https://github.com/inveniosoftware/invenio-communities/issues/620

* use indexer obj when refreshing to ensure that the index name is
  computed adding configured prefix and suffix.


invenio_records_resources/services/uow.py
```diff
@@ -187,14 +187,15 @@ def on_commit(self, uow):
 class IndexRefreshOp(Operation):
     """Search index refresh operation."""
 
-    def __init__(self, index, **kwargs):
+    def __init__(self, indexer, index, **kwargs):
```
ppanero, line 190: The indexer obj has access to the `record_cls` which allows us to already access the index already. See [here](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/records/service.py#L51). I'm 99% sure we can remove the `index` parameter.

Issue #305, opened 2022-03-09
global: remove elastic search 6 support


lnielsen:
ES6 has reached end of life and we can therefore remove support for ES6. 

Search for ``lt_es7`` in the source code to find places with specifics for Elasticsearch 6.

chriz-uniba:
I searched for `lt_es7` in the `inveniosoftware` github-organization and I think, that I found all repositories including `lt_es7`.
However, I think there is even more, when you are searching for `ES_VERSION`.

At this point it gets harder and harder to find 'all' so I will stop here and leave the current PRs for review (whenever time and effort is available).

PR #340, opened 2022-05-16
Removes es6.

chriz-uniba:
Refs https://github.com/inveniosoftware/invenio-records-resources/issues/305

Dependency: https://github.com/inveniosoftware/invenio-vocabularies/pull/193


ppanero:
It was included in the addition of support for opensearch.

Issue #347, opened 2022-06-22
REST-API: Get a record file 's metadata result to 500 INTERNAL SERVER ERROR  if file not exists


Samk13:
Package version (if known): [v0.19.17](https://github.com/inveniosoftware/invenio-records-resources/commit/f1fb9a849d03af1d6ec4cddfc4e140a06788783b)

## Describe the bug
following the [docs](https://inveniordm.docs.cern.ch/reference/rest_api_drafts_records/#:~:text=a%20record%20file%27s-,metadata,-%C2%B6): 
Using Postman:
Providing wrong file name will results to internal server error
GET {{baseURL}}/api/records/{{record-id}}/files/{{unexisting-file-name}}

## Steps to Reproduce
1. Go to Postman
2. Send GET request: {{baseURL}}/api/records/{{record-id}}/files/{{unexisting-file-name}}
3. you get 500 error

## Expected behavior
quit gracefully with an error message

Samk13:
Same error for [Download a record file](https://inveniordm.docs.cern.ch/reference/rest_api_drafts_records/#:~:text=filename%7D%22%0A%20%20%7D%0A%7D-,Download%20a%20record%20file,-%C2%B6):
{{baseURL}}/api/records/{{r-id}}/files/{{file-name}}/content

ntarocco:
Do you have by any chance the stack trace?

Samk13:
Get record meta data:

```console 
Traceback (most recent call last):
  File "/home/samk13/.virtualenvs/rdm-master-venv/lib/python3.9/site-packages/flask/app.py", line 2095, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/samk13/.virtualenvs/rdm-master-venv/lib/python3.9/site-packages/werkzeug/middleware/proxy_fix.py", line 187, in __call__
    return self.app(environ, start_response)
  File "/home/samk13/.virtualenvs/rdm-master-venv/lib/python3.9/site-packages/werkzeug/middleware/dispatcher.py", line 78, in __call__
    return app(environ, start_response)
  File "/home/samk13/.virtualenvs/rdm-master-venv/lib/python3.9/site-packages/flask/app.py", line 2095, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/samk13/.virtualenvs/rdm-master-venv/lib/python3.9/site-packages/werkzeug/middleware/proxy_fix.py", line 187, in __call__
    return self.app(environ, start_response)
[...]

PR #348, opened 2022-06-23
record files: add error response for unknown files

NRodriguezcuellar:
At the moment when requesting from an endpoint that expects to receive a filename (for example `/api/records/{record_id}/files/{filename}`) and the filename does not exist, it will throw an internal server error, it should gracefully return that the resource does not exist. 

Added error catching when a file is being looked up based on it's `key/filename `, if a key error is raised it will return a 404 error to the user.

closes https://github.com/inveniosoftware/invenio-records-resources/issues/347

### DIFF
--- a/invenio_records_resources/services/files/service.py
+++ b/invenio_records_resources/services/files/service.py
@@ -95,6 +95,9 @@ class FileService(Service):
         """Update the metadata of a file."""
         record = self.get_record(id_, identity, "create_files")
 
+        if file_key not in record.files:
+            return None
+
         self.run_components(
             "update_file_metadata", identity, id_, file_key, record, data, uow=uow
         )
@@ -111,6 +114,9 @@ class FileService(Service):
         """Read the metadata of a file."""
         record = self.get_record(id_, identity, "read_files")
 
+        if file_key not in record.files:
+            return None
+
         self.run_components("read_file_metadata", identity, id_, file_key, record)
 
         return self.file_result_item(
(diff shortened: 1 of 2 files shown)


invenio_records_resources/services/files/service.py
```diff
@@ -95,72 +95,84 @@ def update_file_metadata(self, identity, id_, file_key, data, uow=None):
...
+            self.run_components("commit_file", identity, id_, file_key, record, uow=uow)
 
-        return self.file_result_item(
-            self,
-            identity,
-            record.files[file_key],
-            record,
-            links_tpl=self.file_links_item_tpl(id_),
-        )
+            return self.file_result_item(
```
ntarocco, line 167: Shall we simply check here if:
```
file_key in record.files
```
and return None? Same below.

NRodriguezcuellar: I was contemplating the same but i thought that it could also help catch any other keyErrors that may be raised, or allow for future exceptions to be added. What do you think?

zzacharo: I think I would go with what @ntarocco said because catching all KeyErrors is not always acceptable. I.e now you use the KeyError because you anticipate that the file is not there. Then, returning None instructs a special behaviour to be applied in the view/ui. Imagine now that there is another KeyError just because there is a bug somewhere. You will get the same behaviour while hiding the actual bug thus making the debugging process much harder...

Pineirin (approved):

invenio_records_resources/services/files/service.py
```diff
@@ -95,72 +95,85 @@ def update_file_metadata(self, identity, id_, file_key, data, uow=None):
...
+            )
+        else:
+            return None
 
     def read_file_metadata(self, identity, id_, file_key):
         """Read the metadata of a file."""
         record = self.get_record(id_, identity, "read_files")
 
-        self.run_components("read_file_metadata", identity, id_, file_key, record)
+        if file_key in record.files:
```
kpsherva, line 117: minor, could improve readability
```suggestion
        if file_key not in in record.files:
            return None
```

PR #349, opened 2022-06-30
file results: add params

NRodriguezcuellar:
added 2 params which are used to control what and how cache control headers are being set

partly closes https://github.com/inveniosoftware/invenio-communities/issues/718

### DIFF
--- a/invenio_records_resources/services/files/results.py
+++ b/invenio_records_resources/services/files/results.py
@@ -43,10 +43,10 @@ class FileItem(RecordItem):
         """Get links for this result item."""
         return self._links_tpl.expand(self._file)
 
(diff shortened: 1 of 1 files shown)


invenio_records_resources/services/files/results.py
```diff
@@ -43,10 +43,10 @@ def links(self):
         """Get links for this result item."""
         return self._links_tpl.expand(self._file)
 
-    def send_file(self, restricted=True, as_attachment=False):
+    def send_file(self, restricted=True, as_attachment=False, max_age=None, no_cache=False):
```
ntarocco, line 46: See comments in other PRs, not sure it is needed.

Issue #365, opened 2022-07-22
BaseCF: add `required`, `validation` and `custom_errors`


ppanero:
- https://github.com/inveniosoftware/invenio-records-resources/pull/371

ppanero:
PR has been merged to feature/custom-fields branch

PR #369, opened 2022-07-26
remove custom field service

zzacharo:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1011


invenio_records_resources/services/custom_fields/__init__.py
```diff
@@ -8,15 +8,12 @@
 """Custom Fields for InvenioRDM."""
```
ppanero, line 9: discussion, not blocking:

it is now still service level? Or now is records (data) layer? could be a separate one?

ppanero (approved):

invenio_records_resources/services/custom_fields/__init__.py
zzacharo: maybe more of a data layer but let's take it when we tackle https://github.com/inveniosoftware/invenio-records-resources/issues/365 !

PR #370, opened 2022-07-28
custom fields: update schema fields

ppanero:


constraints-dev.txt
```diff
@@ -0,0 +1 @@
+git+https://github.com/inveniosoftware/invenio-records.git@feature/custom-fields
```
zzacharo, line 1: Maybe we add a comment that this file is used only when a PR is targeting the `feature/branch`

.github/workflows/tests-dev.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Generate dependencies
+        run: |
+          pip install wheel requirements-builder
+          requirements-builder -e "$EXTRAS" --level=${{ matrix.requirements-level }} setup.py > .${{ matrix.requirements-level }}-${{ matrix.python-version }}-requirements.txt
```
zzacharo, line 59: I think this should be changed as we don't want to install dev releases in our external dependencies but rather install the `constraints-dev.txt`

.github/workflows/tests-dev.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: Cache pip
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('.${{ matrix.requirements-level }}-${{ matrix.python-version }}-requirements.txt') }}
+
+      - name: Install dependencies
+        run: |
+          pip install -r .${{ matrix.requirements-level }}-${{ matrix.python-version }}-requirements.txt -c constraints-${{ matrix.requirements-level }}.txt
```
zzacharo, line 69: We can maybe keep the `requirements-level` to `pypi` but hardcode or separate the constraint level to `dev`. Can you extend yml files? maybe we could reuse most of the stuff already in `tests.yml` and configure/override the needed parts

Issue #358, opened 2022-07-19
custom fields: can `relation_cls` be removed?


ppanero:
Currently we are using this attribute [relation_cls](https://github.com/inveniosoftware/invenio-records-resources/blob/custom-fields/invenio_records_resources/services/custom_fields/base.py#L22) to be able to know when a custom field is a vocabulary and when it's not. Is there another way to avoid having this flag?

It is breaking the separation of concerns. Having a `relation_cls` attr here because of `invenio-vocabularies` generates a "theoretical" circular dependency, as in "it's making records-resources be aware of vocabularies" (even if does not require its installation).

zzacharo:
Moved `relation_cls` to `VocabularyCF` for now. Checks are done on `getattr(cf, 'relation_cls', None)` to check if it's a vocabulary custom field or not

PR #371, opened 2022-08-03
custom fields: support required fields, custom validation and error m…

ppanero:
- closes https://github.com/inveniosoftware/invenio-records-resources/issues/365

### DIFF
--- a/invenio_records_resources/services/custom_fields/base.py
+++ b/invenio_records_resources/services/custom_fields/base.py
@@ -13,10 +13,11 @@ from abc import ABC, abstractmethod
 class BaseCF(ABC):
     """Base Custom Field class."""
 
-    def __init__(self, name):
+    def __init__(self, name, field_args=None):
         """Constructor."""
         self.name = name
         self.relation_cls = None
+        self._field_args = field_args or {}
         super().__init__()
 
     @property
@@ -28,10 +29,10 @@ class BaseCF(ABC):
     @property
     @abstractmethod
     def field(self):
-        """Marshmallow schema for vocabulary custom fields."""
+        """Marshmallow field for custom fields."""
         pass
 
     @property
     def ui_field(self):
-        """Marshmallow UI schema for vocabulary custom fields."""
-        return self.schema()
+        """Marshmallow UI field for custom fields."""
+        return self.field()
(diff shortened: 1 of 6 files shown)


invenio_records_resources/services/custom_fields/base.py
```diff
@@ -28,10 +41,13 @@ def mapping(self):
     @property
     @abstractmethod
     def field(self):
-        """Marshmallow schema for vocabulary custom fields."""
+        """Marshmallow schema for vocabulary custom fields.
```
zzacharo, line 44: We need to update the docstring

invenio_records_resources/services/custom_fields/base.py
```diff
@@ -28,10 +41,13 @@ def mapping(self):
...
+
+        Must call _field_kwargs() to get the kwargs for the field.
+        """
         pass
 
     @property
     def ui_field(self):
         """Marshmallow UI schema for vocabulary custom fields."""
```
zzacharo, line 52: same here, although this property needs to be checked if it's a field or a schema

invenio_records_resources/services/custom_fields/base.py
```diff
@@ -13,12 +13,25 @@
 class BaseCF(ABC):
     """Base Custom Field class."""
 
-    def __init__(self, name):
+    def __init__(self, name, required=False, validate=None, error_messages=None):
```
zzacharo, line 16: WHat about accepting general `field_kwargs` e.g:
```suggestion
    def __init__(self, name, field_kwargs=None):
```
What's the behaviour when you pass non existing kwargs in marshmallow field?

invenio_records_resources/services/custom_fields/base.py
```diff
@@ -13,12 +13,25 @@
...
         """Constructor."""
         self.name = name
         self.relation_cls = None
```
zzacharo, line 19: I think this can be moved in another base class eg: `BaseRelationCF` where all the custom fields related to another class i.e Vocabulary can implement their specific API...wdyt?

ppanero: While I fully agree, it is not possible right now because that attribute is used in many parts of the custom fields implementation to differentiate normal CF from Vocabulary CFs. Solving it, is a wider topic that will be tackled here <https://github.com/inveniosoftware/invenio-records-resources/issues/358>

zzacharo (approved):

Issue #356, opened 2022-07-14
custom fields: ensure mapping compatibility with OpenSearch


ppanero:
The `mapping` property of the `CF` classes is returning a dictionary with the ES configuration. However, the backend can change in the future (e.g. to OpenSearch). Adding a layer of abstration would provide transparent compatibility.

This layer would mean having e.g. `Keyword`, `Text` classes. This could be hosted in invenio-search provided they don't exist in elasticsearch-dsl, elasticsearch-py or the corresponding amazon forks.

ppanero:
PR https://github.com/inveniosoftware/invenio-records-resources/pull/372

ppanero:
Regarding the `TextCF`:
> why not always adding the keyword by default?
There is also this [ignore_above](https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html) which might be useful...

See https://github.com/inveniosoftware/invenio-records-resources/pull/377#discussion_r949996829

ppanero:
This was implemented.

PR #372, opened 2022-08-04
custom fields: use elasticsearch field abstractions

ppanero:
- closes https://github.com/inveniosoftware/invenio-records-resources/issues/356

WIP: Would be nice to not need the `to_dict`. Creating a layer of abstraction (in invenio-search?) for this would make later changes due to compatibility/opensearch migrations transparent at this level.


ppanero:
It was decided to block the PR until the compatibility layer has been implemented (merged) in the other layers. Then the extra layer in invenio-search will be implemented.

ppanero:
closed in favour of Mapping clases

PR #387, opened 2022-09-09
custom fields: edtf and iso date support

### LINKED ISSUE #385: EDTF CF

ppanero:
- closes #385

### DIFF
--- a/invenio_records_resources/services/custom_fields/date.py
+++ b/invenio_records_resources/services/custom_fields/date.py
@@ -1,0 +1,39 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Custom Fields for InvenioRDM."""
+
+from marshmallow import fields
+from marshmallow_utils.fields import EDTFDateString, ISODateString
+
+from .base import BaseListCF
+
+
+class ISODateStringCF(BaseListCF):
+    """ISO date custom field."""
+
+    def __init__(self, name, **kwargs):
+        """Constructor."""
+        super().__init__(name, field_cls=ISODateString, **kwargs)
+
+    @property
+    def mapping(self):
+        """Return the mapping."""
+        return {"type": "date"}
+
+
+class EDTFDateStringCF(BaseListCF):
+    """EDTF date custom field."""
+
+    def __init__(self, name, **kwargs):
+        """Constructor."""
+        super().__init__(name, field_cls=EDTFDateString, **kwargs)
+
+    @property
+    def mapping(self):
+        """Return the mapping."""
+        return {"type": "date", "fields": {"date_range": {"type": "date_range"}}}
--- a/invenio_records_resources/services/custom_fields/date.py
+++ b/invenio_records_resources/services/custom_fields/date.py
@@ -7,8 +7,14 @@
 
 """Custom Fields for InvenioRDM."""
 
+import calendar
+
+from arrow import Arrow
+from edtf import parse_edtf
+from edtf.parser.edtf_exceptions import EDTFParseException
 from marshmallow import fields
 from marshmallow_utils.fields import EDTFDateString, ISODateString
+from pytz import utc
 
 from .base import BaseListCF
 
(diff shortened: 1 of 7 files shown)


invenio_records_resources/services/custom_fields/date.py
```diff
@@ -0,0 +1,99 @@
...
+    @property
+    def field(self):
+        """Marshmallow field custom fields."""
+        _schema = EDTFDateString(**self._field_args)
+        if self._multiple:
+            return fields.List(_schema)
+        return _schema
+
+    @classmethod
+    def _format_date(cls, date):
```
ppanero, line 61: comment: quite some duplication of code with `EDTFDumperExt` (implemented in rdm-records). Could it be moved here and re-used?

ppanero:
Tested in an instance with the following config:

```python

# Custom fields
from invenio_records_resources.services.custom_fields import EDTFDateStringCF, ISODateStringCF

RDM_CUSTOM_FIELDS = [
    EDTFDateStringCF(name="edtf"),
    ISODateStringCF(name="iso"),
]

RDM_CUSTOM_FIELDS_UI = [
    {
        "section": _("CERN Experiment"),
        "fields": [
            dict(
                field="edtf",  # this should be validated against the defined fields in `RDM_CUSTOM_FIELDS`
                ui_widget="Input",  # user defined widget in my-site
                props=dict(
                    label="EDTF Date",
                    placeholder="Type...",
                    icon="lab",
                    description="You should fill this...",
                )
            ),
            dict(
                field="iso",  # this should be validated against the defined fields in `RDM_CUSTOM_FIELDS`
                ui_widget="Input",  # user defined widget in my-site
                props=dict(
                    label="EDTF Date",
                    placeholder="Type...",
                    icon="lab",
                    description="You should fill this...",
                )
            ),
        ]
    }
]
```

Mapping result after services setup:
```
"custom_fields": {
    "properties": {
        "edtf": {
            "properties": {
                "date": {
                    "type": "date"
                },
                "date_range": {
                    "type": "date_range"
                }
            }
        },
        "iso": {
            "type": "date"
        }
    }
},
```

invenio_records_resources/services/custom_fields/date.py
```diff
@@ -0,0 +1,101 @@
...
+                record[cf_key][self.name] = self._calculate_date_range(date)
+
+    def load(self, record, cf_key="custom_fields"):
+        """Load the custom field.
+
+        Gets both the record and the custom fields key as parameters.
+        This supports the case where a field is based on others, both
+        custom and non-custom fields.
+        """
+        date = record.get(cf_key, {}).pop(self.name, None)
```
ppanero, line 99: Not sure why, but on search a record passes twice by the dumper, the second time without `metadata` and `custom_fields`.  That's why the first is a `.get()` instead of `[]`
```shell
(Pdb) record.keys()
dict_keys(['parent', 'versions', 'has_draft', '$schema', 'pid', 'id', 'files', 'metadata', 'custom_fields', 'pids', 'access', 'uuid', 'version_id', 'created', 'updated'])
(Pdb) record.keys()
dict_keys(['id', 'pid', 'access', '$schema', 'communities', 'uuid', 'version_id', 'created', 'updated'])
```

invenio_records_resources/services/custom_fields/date.py
zzacharo: that would be useful indeed! Is it quick to move?

tests/services/custom_fields/test_date_cf.py
```diff
@@ -0,0 +1,60 @@
...
+        "properties": {
+            "date": {"type": "keyword"},
+            "date_range": {"type": "date_range"},
+        },
+    }
+
+
+def test_edtfdatestring_validate():
+    cf = EDTFDateStringCF("name")
+    schema = Schema.from_dict({"test": cf.field})()
```
zzacharo, line 52: **shelvable**: it seems to me the tests are testing the edtf main code while this has been tested or as you mentioned before, we should reuse what was already implemented in rdm-records... Myabe we should actually test the dump/load in a record-oriented workflow I.e create/update/search...

zzacharo (approved):

tests/services/custom_fields/test_date_cf.py
ppanero: I did part of _i.e create/update/search..._, checking both ES, API and DB contents. But I fully agree that's manual testing... I dont think reusing code is a fast move because of different layers. I would open an issue for that.

invenio_records_resources/services/custom_fields/date.py
ppanero: Opened #394

tests/services/custom_fields/test_date_cf.py
ppanero: Added a test see `test_dumpers.py`

PR #388, opened 2022-09-09
custom fields: simplify implementation of multiple values (list)

### LINKED ISSUE #386: can we avoid re-implementing fields.list with every CF class? move that to the actual BaseListCF
Something like:

```python

class BaseListCF:
    
    def field(self, field):
          if multiple:
                return field.List(field)


class TextCF:
      
     def field(self):
         super.field(SanitizedUnicode)
```
     
Needs thinking since `field` is a property not a function.

ppanero:
- closes #386 

This enables removing duplicated code. See #387, where both date fields need to reimplement the logic of `if multiple...`. Now it gets delegated to the `BaseListCF` class.

### DIFF
--- a/invenio_records_resources/services/custom_fields/base.py
+++ b/invenio_records_resources/services/custom_fields/base.py
@@ -37,7 +37,7 @@ class BaseCF(ABC):
         return self.field
 
 
-class ListBaseCF(BaseCF):
+class BaseListCF(BaseCF):
     """Base Custom Field class."""
 
     def __init__(self, name, multiple=False, **kwargs):
--- a/invenio_records_resources/services/custom_fields/base.py
+++ b/invenio_records_resources/services/custom_fields/base.py
@@ -9,6 +9,8 @@
 
 from abc import ABC, abstractmethod
 
+from marshmallow.fields import List
+
 
 class BaseCF(ABC):
     """Base Custom Field class."""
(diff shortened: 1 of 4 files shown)


invenio_records_resources/services/custom_fields/base.py
```diff
@@ -37,10 +39,18 @@ def ui_field(self):
...
+    def __init__(self, name, field_cls, multiple=False, **kwargs):
         """Constructor."""
         super().__init__(name, **kwargs)
         self._multiple = multiple
+        self._field_cls = field_cls
+
+    @property
+    def field(self):
+        """Marshmallow field custom fields."""
+        _field = self._field_cls(**self._field_args)
```
zzacharo, line 54: I think here you could do instead 
```suggestion
        _field = self.__class__(**self._field_args)
```
it should give the actual instance class....

ppanero: but `__class__` would point to `BaseCF` or `BaseListCF`. While in this case we are talking about Marshmallow field classes (e.g. `SanitizeUnicode`)

zzacharo: my bad! I was confused... Maybe we can add to the constrcutor's docstring the definition of the parameters, so it's all clear and tidy :)

ppanero: Will be tidy, not sure will ever be "clear" with all naming issues 😂

zzacharo (approved):

PR #390, opened 2022-09-09
facets: add support for cf nested terms

### LINKED ISSUE #389: create CF specific facet for NestedTermsFacet
https://github.com/inveniosoftware/invenio-records-resources/pull/377/files#r949997222

This is needed for the case of custom complex fields coded by users.

**Update**: as per comment, if the path passed to the `CFTermsFacet` is a dot separated string then the nested facet works.

What is missing is:

- [x] Create a CFNestedTermsFacet

ppanero:
- closes #389 

# ⚠️ Missing tests. We should add also for the previous `CFTermsFacet`

### DIFF
--- a/invenio_records_resources/services/records/facets/facets.py
+++ b/invenio_records_resources/services/records/facets/facets.py
@@ -254,7 +254,16 @@ class NestedTermsFacet(TermsFacet):
         return ret_val
 
 
-class CFTermsFacet(TermsFacet):
+class CFFacetMixin:
+    """Mixin to abstract the custom fields path."""
+
+    @classmethod
+    def field(cls, field):
+        """Format field with `custom_fields` prefix."""
+        return f"custom_fields.{field}"
+
+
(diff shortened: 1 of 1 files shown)


invenio_records_resources/services/records/facets/facets.py
```diff
@@ -266,7 +275,15 @@ def __init__(self, field=None, label=None, value_labels=None, **kwargs):
...
-        return f"custom_fields.{field}"
+
+class CFNestedTermsFacet(CFFacetMixin, NestedTermsFacet):
+    """Nested Terms facet for custom fields.
+
+    Works exactly as NestedTermsFacet except that prepends the prefix `custom_fields`
+    in the field definition.
+    """
+
+    def __init__(self, field=None, label=None, value_labels=None, **kwargs):
```
ppanero, line 286: question: can we also move the `__init__` to the constructor and play with MRO so both `CFTermsFacet` and `CFNestedFacet` are just a one liner?

zzacharo (approved):

PR #392, opened 2022-09-12
custom fields: add search version parameter to mapping method

### LINKED ISSUE #391: CF: provide search engine info to `mapping` property call to support returning different mappings.

ppanero:
- closes #391

### DIFF
--- a/invenio_records_resources/services/custom_fields/base.py
+++ b/invenio_records_resources/services/custom_fields/base.py
@@ -21,12 +21,6 @@ class BaseCF(ABC):
 
     @property
     @abstractmethod
-    def mapping(self):
-        """Return the mapping."""
-        pass
-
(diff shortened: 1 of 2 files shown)


invenio_records_resources/services/custom_fields/base.py
```diff
@@ -36,6 +30,13 @@ def ui_field(self):
         """Marshmallow UI field for custom fields."""
         return self.field
 
+    @abstractmethod
+    def mapping(self, search_version=None):
```
ppanero, line 34: comment: default to `None` seemed better than defaulting to ES or OS. Then the default can be set at invenio-cli (according to what was chosen when bootstrapping the instance).

ppanero:
closed in favour of #396

PR #396, opened 2022-09-14
custom fields: abstract mapping definitions

### LINKED ISSUE #391: CF: provide search engine info to `mapping` property call to support returning different mappings.

ppanero:
- closes #391


invenio_records_resources/services/custom_fields/mappings.py
```diff
@@ -0,0 +1,64 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio-Records-Resources is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Custom fields mappings."""
```
ppanero, line 9: Note: the search engine and version information could be imported here (i.e. `SEARCH_DISTRIBUTION` from invenio-search) and used to calcualte based on it.

zzacharo (approved):

invenio_records_resources/services/custom_fields/mappings.py
```diff
@@ -0,0 +1,64 @@
...
+
+from abc import abstractmethod
+
+
+class Mapping:
+    """Base class for mappings."""
+
+    @abstractmethod
+    def mapping(self):
```
slint, line 18:
```suggestion
    def to_dict(self):
```

got weirded out at some point :)

slint (approved):
Looks good as a solution, maybe worth looking into how it's also done in ElasticsearchDSL (which we were also discussing about being a possible "compatibility" layer for defining mappings programatically): https://elasticsearch-dsl.readthedocs.io/en/latest/persistence.html#data-types

ppanero:
@slint the main issue on using ElasticsearchDSL is that is a library that it's exactly the point where it would differ. See [https://github.com/inveniosoftware/invenio-search/pull/216/files#diff-fa602a8a75dc9dcc92261bac5f533c2a85e34fcceaff63b3a3a81d9acde2fc52R44]), OpenSearch would use a different library, so we would still need an extra layer.

slint:
> @slint the main issue on using ElasticsearchDSL is that is a library that it's exactly the point where it would differ. See [https://github.com/[inveniosoftware/invenio-search/pull/216](https://github.com/inveniosoftware/invenio-search/pull/216)/files#diff-fa602a8a75dc9dcc92261bac5f533c2a85e34fcceaff63b3a3a81d9acde2fc52R44]), OpenSearch would use a different library, so we would still need an extra layer.

I'm sorry should have clarified, it's more about the "style" that fields/classes are defined with and what kind of parameters they accept. E.g. we're now passing `_use_as_filter` all the way down (which in our case means "also add a `keyword` sub-field"), but maybe this could be done on the level of `TextCF.mapping` property and not be handled by the `mappings.TextMapping` class.

PR #406, opened 2022-09-29
search: remove suggestions support ES<7

ntarocco:


ppanero (approved):
LGTM! There are some comments in line 50 that could be removed too.

Issue #410, opened 2022-10-03
files (sciebo): make allowed external domains for file fetchning configurable


ppanero:

ppanero:
Closing this to avoid confusion. Will be implemented at #416

Issue #408, opened 2022-10-03
files (sciebo): add support to fetch files from links payload


ppanero:
- [x] define how the payload would look like (data model) for files, and how it would be modified once the files would be fetched.
- [x] implement async (celery) task to fetch files (take into account queue names)
    - [x] should re-index record to which the files belong to with appropriate information, how do we handle concurrency? how do we avoid re-launching tasks exponentially?

Issue #409, opened 2022-10-03
files (sceibo): fetching error handling


ppanero:
- The fetched file is corrupted
- Timeouts
- Slow downloads
- Connection's cut
- Access token expiration
- etc.

ppanero:
Also see [comment](https://github.com/inveniosoftware/invenio-records-resources/pull/416#discussion_r1001871066) related to file size.

PR #416, opened 2022-10-11
services: implement file fetching from external storage

ppanero:
- closes https://github.com/inveniosoftware/invenio-records-resources/issues/408
- closes https://github.com/inveniosoftware/invenio-records-resources/issues/410

Now there will be two "workflows" to upload files:
a) if the payload **does not** contain a `uri`/`sotrage_class==E` it goes as before: `init` will create the file record (no object version nor file instance), Then requests to `/content` and `/commit` are needed from the user to get the file in.

b) if the payload **does** contain a `uri`/`sotrage_class==E`, the `init` creates an object version and a file instance, and triggers the task to do a `set_content` and `commit`. Those actions are forbiden by the user.

### DIFF
--- a/invenio_records_resources/records/systemfields/files/manager.py
+++ b/invenio_records_resources/records/systemfields/files/manager.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2020-2021 CERN.
+# Copyright (C) 2020-2022 CERN.
 # Copyright (C) 2020-2021 Northwestern University.
 #
 # Invenio-Records-Resources is free software; you can redistribute it and/or
@@ -60,8 +60,7 @@ from collections.abc import MutableMapping
 from functools import wraps
 
 from invenio_files_rest.errors import InvalidKeyError, InvalidOperationError
-from invenio_files_rest.models import Bucket, ObjectVersion
-from invenio_records.systemfields import SystemField
+from invenio_files_rest.models import Bucket, FileInstance, ObjectVersion
 
 
 def ensure_enabled(func):
@@ -145,7 +144,7 @@ class FilesManager(MutableMapping):
 
     # TODO: "create" and "update" should be merged somehow...
     @ensure_enabled
-    def create(self, key, obj=None, stream=None, data=None):
+    def create(self, key, obj=None, stream=None, data=None, **kwargs):
         """Create/initialize a file."""
         assert not (obj and stream)
 
@@ -154,8 +153,12 @@ class FilesManager(MutableMapping):
 
         rf = self.file_cls.create({}, key=key, record_id=self.record.id)
         if stream:
-            obj = ObjectVersion.create(self.bucket, key, stream=stream)
+            obj = ObjectVersion.create(self.bucket, key, stream=stream, **kwargs)
         if obj:
+            if isinstance(obj, dict):
+                fi = FileInstance.create()
+                fi.set_uri(**obj.pop("file"))
+                obj = ObjectVersion.create(self.bucket, key, fi.id)
             rf.object_version_id = obj.version_id
             rf.object_version = obj
         if data:
@@ -165,15 +168,23 @@ class FilesManager(MutableMapping):
         return rf
 
     @ensure_enabled
-    def update(self, key, obj=None, stream=None, data=None):
-        """Update a file."""
-        assert not (obj and stream)
+    def create_obj(self, key, stream, data=None, **kwargs):
+        """Create an ObjectVersion but do not pop it to the top of the stack."""
+        # this is used on set_file_content, since the file is not yet commited
         rf = self.get(key)
         if rf is None:
             raise InvalidKeyError(description=f"File with {key} does not exist.")
 
+        return ObjectVersion.create(self.bucket, key, stream=stream, **kwargs)
+
+    @ensure_enabled
+    def update(self, key, obj=None, stream=None, data=None, **kwargs):
+        """Update a file."""
+        assert not (obj and stream)
+        rf = self.get(key)
+
         if stream:
-            obj = ObjectVersion.create(self.bucket, key, stream=stream)
+            obj = self.create_obj(key, stream=stream, **kwargs)
         if obj:
             rf.object_version_id = obj.version_id
             rf.object_version = obj
@@ -181,6 +192,15 @@ class FilesManager(MutableMapping):
             rf.metadata = data
             rf.commit()
         return rf
+
+    @ensure_enabled
+    def commit(self, file_key):
+        """Commit a file."""
+        # TODO: Add other checks here (e.g. verify checksum, S3 upload)
+        file_obj = ObjectVersion.get(self.bucket.id, file_key)
+        if not file_obj:
+            raise Exception(f"File with key {file_key} not uploaded yet.")
+        self[file_key] = file_obj
 
     @ensure_enabled
     def delete(self, key, remove_obj=True, softdelete_obj=False):
--- a/invenio_records_resources/services/files/components/metadata.py
+++ b/invenio_records_resources/services/files/components/metadata.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2021 CERN.
+# Copyright (C) 2021-2022 CERN.
 #
 # Invenio-Records-Resources is free software; you can redistribute it and/or
 # modify it under the terms of the MIT License; see LICENSE file for more
@@ -10,9 +10,8 @@
 
 from copy import deepcopy
 
-from invenio_files_rest.models import ObjectVersion
-
 from ..schema import InitFileSchema
+from ..transfer import Transfer
 from .base import FileServiceComponent
 
 
@@ -25,17 +24,18 @@ class FileMetadataComponent(FileServiceComponent):
         validated_data = schema.load(data)
         for file_metadata in validated_data:
             temporary_obj = deepcopy(file_metadata)
-            record.files.create(temporary_obj.pop("key"), data=temporary_obj)
+            file_type = temporary_obj.pop("storage_class", None)
+            transfer = Transfer.get_transfer(
+                file_type, service=self.service, uow=self.uow
+            )
+            _ = transfer.init_file(record, temporary_obj)
 
     def update_file_metadata(self, identity, id, file_key, record, data):
         """Update file metadata handler."""
+        # FIXME: move this call to a transfer call
         record.files.update(file_key, data=data)
 
     # TODO: `commit_file` might vary based on your storage backend (e.g. S3)
     def commit_file(self, identity, id, file_key, record):
         """Commit file handler."""
-        # TODO: Add other checks here (e.g. verify checksum, S3 upload)
-        file_obj = ObjectVersion.get(record.bucket.id, file_key)
-        if not file_obj:
-            raise Exception(f"File with key {file_key} not uploaded yet.")
-        record.files[file_key] = file_obj
+        Transfer.commit_file(record, file_key)
--- a/invenio_records_resources/services/files/generators.py
+++ b/invenio_records_resources/services/files/generators.py
@@ -1,0 +1,47 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio-Records-Resources is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""File permissions generators."""
+
+
+from invenio_access.permissions import any_user, system_process
+from invenio_records_permissions.generators import Generator
+from invenio_search.engine import dsl
+
+from .transfer import TransferType
+
+
+class AnyUserIfFileIsLocal(Generator):
+    """Allows any user."""
+
+    def needs(self, **kwargs):
+        """Enabling Needs."""
+        record = kwargs["record"]
+        file_key = kwargs.get("file_key")
+        is_file_local = True
+        if file_key:
+            file_record = record.files.get(file_key)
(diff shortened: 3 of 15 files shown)


invenio_records_resources/records/api.py
```diff
@@ -19,6 +20,20 @@
 from invenio_records.systemfields.model import ModelField
 
 
+class FileType(str, Enum):
```
ppanero, line 23: This should be changed for a dynamic enum _a la request event type_

invenio_records_resources/records/api.py
```diff
@@ -115,6 +130,27 @@ def record(self):
     record_id = ModelField()
     _record = ModelField("record", dump=False)
 
+    def is_external(self):
```
ppanero, line 133: Move this methods to the `File` wrapper class?

invenio_records_resources/records/systemfields/files/manager.py
```diff
@@ -156,6 +157,12 @@ def create(self, key, obj=None, stream=None, data=None):
         if stream:
             obj = ObjectVersion.create(self.bucket, key, stream=stream)
         if obj:
+            if isinstance(obj, dict):
```
ppanero, line 160: Not the cleanest of the ways. Accessing DB models at this level is necessary due to the strictness of both `ObjectVersion` and `FileInstance` _create_ methods. 

It should be evaluated if this is "acceptable" enough. Afterall, there is no `File API` class (i.e. and `api.py` in files-rest). Therefore, this system field acts as one.

ppanero (commented):
.

invenio_records_resources/services/files/transfer.py
```diff
@@ -0,0 +1,157 @@
...
+    def __init__(self, type):
+        """Constructor."""
+        self.type = type
+
+    @abstractmethod
+    def init_file(self, record, file_metadata):
+        """Initialize a file."""
+        pass
+
+    def set_file_content(self, record, file, file_key, stream, content_length):
```
ppanero, line 52: TODO: create generator to disallow set by users when file==external, allowing only system identity (tasks)

invenio_records_resources/services/files/generators.py
```diff
@@ -0,0 +1,46 @@
...
+"""File permissions generators."""
+
+
+from invenio_access.permissions import any_user, system_process
+from invenio_records_permissions.generators import Generator
+from invenio_search.engine import dsl
+
+
+# TODO: Naming
+class AnyUserIfFileIsLocal(Generator):
```
ppanero, line 18: In general, we need (and will more and more) permissions over more fine grained file actions. However, I'm not convinced about making the generator aware of `S` and `E` (Storage classes enum). On the other hand, I'm not sure how to abstract it except having "can_<action>" functions in the `Transfer` object... (seems like a permission duplicate).

invenio_records_resources/services/files/service.py
```diff
@@ -48,12 +48,12 @@ def check_permission(self, identity, action_name, **kwargs):
         action_name = self.config.permission_action_prefix + action_name
         return super().check_permission(identity, action_name, **kwargs)
 
-    def get_record(self, id_, identity, action):
+    def get_record(self, id_, identity, action, **kwargs):
```
ppanero, line 51: passing kwargs (`file_key`) is unavoidable if we want fine grained file permissions (i.e. depending on file attributes)

tests/mock_module/permissions.py
```diff
@@ -23,6 +25,9 @@ class PermissionPolicy(RecordPermissionPolicy):
     can_update = [AnyUser(), SystemProcess()]
     can_delete = [AnyUser(), SystemProcess()]
     can_create_files = [AnyUser(), SystemProcess()]
+    can_set_content_files = [AnyUserIfFileIsLocal(), SystemProcess()]
```
ppanero, line 28: TODO: if we go with this, we will need to add it in the rdm-permission policy.

invenio_records_resources/services/files/transfer.py
```diff
@@ -0,0 +1,156 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio-Records-Resources is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Files transfer."""
```
ppanero, line 9: I'm not a 100% convinced about this whole implementeation. It's nice to have it abstracted and with a "factory" to decide the call. However, it feels that the complexity was just _moved to another class_.

invenio_records_resources/services/files/transfer.py
```diff
@@ -0,0 +1,156 @@
...
+        """Set file content."""
+        bucket = record.bucket
+        size_limit = bucket.size_limit
+        if content_length and size_limit and content_length > size_limit:
+            desc = (
+                "File size limit exceeded."
+                if isinstance(size_limit, int)
+                else size_limit.reason
+            )
+            raise FileSizeError(description=desc)
```
alejandromumo, line 62: From the point of view of the  user, what would happen if the file is too big? Is this "catchable" so the user can be immediately notified? Without something like messaging I am not sure what would happen since this will run in a task (I suppose?).

invenio_records_resources/services/files/components/metadata.py
```diff
@@ -25,7 +26,18 @@ def init_files(self, identity, id, record, data):
         validated_data = schema.load(data)
         for file_metadata in validated_data:
             temporary_obj = deepcopy(file_metadata)
-            record.files.create(temporary_obj.pop("key"), data=temporary_obj)
+            file_type = temporary_obj.pop("storage_class", None)
+            transfer = Transfer.get_transfer(file_type)
+            file_record = transfer.init_file(record, temporary_obj)
+            if Transfer.fetch_file(file_record.file):
```
alejandromumo, line 32: Small note about wording. 

When I first read this, `fetch_file` looked to me like a function that would fetch the file. However, after reading the implentation, I understand that it's a validation on whether the file is fetchable from an external source (without validating specifically for file type here).

Perhaps something like `is_____()` would be easier to read?

invenio_records_resources/services/files/transfer.py
ppanero: That is a good point, I will add it to the [error handling issue](https://github.com/inveniosoftware/invenio-records-resources/issues/409). 

On the spot ideas:

a) Notify the user when they make the upload: For this we would need to make the `size` parameter mandatory and trust it's veracity or perform a `HEAD` request.
b) Alternatively, since we would not be able to know the file size in advance we could show the error in the deposit form. How would we persist them?

We can discuss this in depth when tackling the above mentioned issue?

invenio_records_resources/services/files/components/metadata.py
ppanero: Agree, or something like `needs_...`.

invenio_records_resources/services/files/transfer.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def __init__(self):
+        """Constructor."""
+        super().__init__(TransferType.LOCAL)
+
+    def init_file(self, record, file_metadata):
+        """Initialize a file."""
+        uri = file_metadata.pop("uri", None)
+        if uri:
+            raise Exception("Cannot set URI for local files.")
```
slint, line 90: minor: probably we'll need a base `TransferException`, but maybe as a follow-up task in error handling.

tests/mock_module/permissions.py
```diff
@@ -23,6 +25,9 @@ class PermissionPolicy(RecordPermissionPolicy):
...
+    can_get_content_files = [AnyUserIfFileIsLocal(), SystemProcess()]
+    can_commit_files = [AnyUserIfFileIsLocal(), SystemProcess()]
     can_read_files = [AnyUser(), SystemProcess()]
     can_update_files = [AnyUser(), SystemProcess()]
-    can_delete_files = [AnyUser(), SystemProcess()]
+    can_delete_files = [AnyUserIfFileIsLocal(), SystemProcess()]
```
slint, line 33: minor: this might not be needed actually, since that could be a way to "cancel" the background upload?

slint (commented):
LGTM! I think we're almost there, just a couple of naming and placement things to avoid data/code migrations in the future.

invenio_records_resources/services/files/transfer.py
slint: Overall, I like the complexity/control move since it also helps with extending transfer types easily. What's a bit "obscure" is that this is called from within a component vs. being core logic of the FilesService.

invenio_records_resources/services/files/service.py
ppanero: Agreed, does not seem to be used anywhere else (e.g. outside of services, where it would be okay to be `_`ed)

invenio_records_resources/services/files/schema.py
```diff
@@ -42,23 +44,57 @@ class Meta:
...
+    checksum = Str()
+    size = Integer()
+
+    @validates("uri")
+    def validate_names(self, value):
+        """Validate the domain of the URI is allowed."""
+        # checking if storage class and uri are compatible is a
+        # business logic concern, not a schema concern.
+        if value:
+            domain = urlparse(value).netloc
```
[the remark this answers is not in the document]
ppanero: Fixed, also added a test (in services, since there is no full schema testing)

invenio_records_resources/services/files/components/metadata.py
```diff
@@ -34,8 +46,4 @@ def update_file_metadata(self, identity, id, file_key, record, data):
     # TODO: `commit_file` might vary based on your storage backend (e.g. S3)
     def commit_file(self, identity, id, file_key, record):
         """Commit file handler."""
-        # TODO: Add other checks here (e.g. verify checksum, S3 upload)
-        file_obj = ObjectVersion.get(record.bucket.id, file_key)
-        if not file_obj:
-            raise Exception(f"File with key {file_key} not uploaded yet.")
-        record.files[file_key] = file_obj
+        record.files.commit(file_key)
```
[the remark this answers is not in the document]
ppanero: I'm still in favour of delegating the final commit to the manager, because otherwise we would have to use `ObjectVersions` and other data layer objects here. On the other hand, I also agree that it should pass by the `Transfer` object in order to make the behaviour configurable and storage-class dependant. I've refactored it to reflect this. (still pending to see what to do with `files.update`, all the other commands follow `service -> cmp -> transfer -> manager`

invenio_records_resources/services/files/components/metadata.py
ppanero: I have move it inside the `FetchTransfer` object. Passing the `uow` is still not optimal. My gut feeling tells me that we will need it for other transfer types, and most important that way we do not expose the specific behaviour outside of the Transfer type.

invenio_records_resources/services/files/schema.py
```diff
@@ -42,23 +44,57 @@ class Meta:
...
+            allowed_domains = current_app.config.get(
+                "RECORDS_RESOURCES_FILES_ALLOWED_DOMAINS"
+            )
+            if domain not in allowed_domains:
+                raise ValidationError("Domain not allowed", field_name="uri")
+
+    @pre_dump(pass_many=False)
+    def fields_from_file_obj(self, data, **kwargs):
+        """Fields coming from the FileInstance model."""
+        # data is a FileRecord instance, might not have a file yet.
```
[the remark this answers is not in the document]
ppanero: Im not sure I understand fully, we are getting them form the `File` wrapper. `data.file` is a `File` object, we are dumping it into the object. We cannot use `data_key` or `attribute` since we need to load from e.g. `uri` and dump from `file.uri`.

ppanero:
Things shelved for the future:
- `service->cmp->transfer` is okay, but still fishy since the transfer object needs the unit of work and service objects (in some cases, e.g. `init_file`. Can this be a core part of the file service? if yes, how does it link to the components? could it become a `TransferComponent` or something like this?
- The generator needs knowledge of `TransferType.LOCAL`, can we avoid this?

invenio_records_resources/services/files/tasks.py
```diff
@@ -0,0 +1,34 @@
...
+
+import requests
+from celery import shared_task
+from invenio_access.permissions import system_identity
+
+from ...proxies import current_service_registry
+
+
+@shared_task(ignore_result=True)
+def fetch_file(service_id, record_id, file_key):
```
slint, line 19: minor: hate myself for missing this one, this should be added as an `invenio_celery.tasks` entrypoint in `setup.cfg`

tests/conftest.py
```diff
@@ -23,6 +23,27 @@
...
+def app_config(app_config):
+    """Override pytest-invenio app_config fixture.
+
+    Needed to set the fields on the custom fields schema.
+    """
+    app_config["RECORDS_RESOURCES_FILES_ALLOWED_DOMAINS"] = [
+        "inveniordm.test",
+    ]
+
+    app_config["FILES_REST_STORAGE_CLASS_LIST"] = {
```
slint, line 36: shelve/comment: we have to remember to configure this in `app-rdm` (and ZenodoRDM)

slint (approved):
We have to redo everything... 



:trollface: 

One missing entrypoint and we're good to go :rocket:

invenio_records_resources/services/files/tasks.py
ppanero: ooops done

slint (approved):

ppanero:
@slint 
- Entry point added (tasks)
- Commits squashes
- Opened issues:
    - add permission generator to rdm-records https://github.com/inveniosoftware/invenio-rdm-records/issues/1119
    - file storage class config https://github.com/inveniosoftware/invenio-app-rdm/issues/1955

PR #419, opened 2022-10-27
Bulk indexing for `rebuild_index`

max-moser:
Make the `service.rebuild_index` method use bulk indexing rather than indexing each item individually.

### DIFF
--- a/invenio_records_resources/services/records/service.py
+++ b/invenio_records_resources/services/records/service.py
@@ -464,11 +464,8 @@ class RecordService(Service):
 
         Note: Skips (soft) deleted records.
         """
-        for rec_meta in self.record_cls.model_cls.query.all():
-            rec = self.record_cls(rec_meta.data, model=rec_meta)
-
-            if not rec.is_deleted:
-                self.indexer.index(rec)
+        records = self.record_cls.model_cls.query.filter_by(is_deleted=False).all()
+        self.indexer.bulk_index([rec.id for rec in records])
 
         return True
 
--- a/invenio_records_resources/services/records/service.py
+++ b/invenio_records_resources/services/records/service.py
@@ -54,10 +54,7 @@ class RecordService(Service):
 
     def record_to_index(self, record):
         """Function used to map a record to an index."""
-        # We are returning "_doc" as document type as recommended by
-        # the search engine documentation. In v8
-        # document types will have been completely removed.
-        return record.index._name, "_doc"
+        return record.index._name
 
     @property
     def schema(self):


kpsherva:
discussion: https://discord.com/channels/692989811736182844/704629170470125578/1035200563392020640

invenio_records_resources/services/records/service.py
```diff
@@ -467,11 +464,8 @@ def rebuild_index(self, identity, uow=None):
 
         Note: Skips (soft) deleted records.
         """
-        for rec_meta in self.record_cls.model_cls.query.all():
-            rec = self.record_cls(rec_meta.data, model=rec_meta)
-
-            if not rec.is_deleted:
-                self.indexer.index(rec)
+        records = self.record_cls.model_cls.query.filter_by(is_deleted=False).all()
+        self.indexer.bulk_index([rec.id for rec in records])
```
ppanero, line 468: The bulk_index method does not take into account the deleted records. Is there a reason for that if to be removed? If not you can .filter() the list to keep the one liner or just split and append only if not deleted.

max-moser: given that the `record.is_deleted` property refers to the model's `is_deleted` flag (from [invenio-records.api.RecordBase](https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py#L105)), the `query.filter_by(is_deleted=False)` should do the same but on an SQL level, no?
if you prefer, we can move this filter back to the list comprehension, i just thought it would be nice to exclude those records already in the query.

invenio_records_resources/services/records/service.py
```diff
@@ -54,10 +54,7 @@ def indexer(self):
 
     def record_to_index(self, record):
         """Function used to map a record to an index."""
-        # We are returning "_doc" as document type as recommended by
-        # the search engine documentation. In v8
-        # document types will have been completely removed.
-        return record.index._name, "_doc"
+        return record.index._name
```
zzacharo, line 57: is this a breaking change?

max-moser: no, not really... it wasn't required in ES for a while now (certainly not in ES7, which is the minimum supported version in invenio nowadays), and is actually incompatible with OSv2.
`invenio-indexer` actually still handles [both cases](https://github.com/inveniosoftware/invenio-indexer/commit/0f28228e9f0700307c7d9574f68acd4ff87d1599#diff-3f68cfb18ee1c14da652ccebd99987da87e79b89c87fa6d5bb75665eee508a13R102), either returning only one value or a tuple – however, in the latter case it issues a warning, which this change fixes.

invenio_records_resources/services/records/service.py
kpsherva: I think the point is to also reindex the deleted records, but please correct me if I am wrong @ppanero

ppanero: @max-moser, you are correct I missed the `filter_by`. Should be good. There is not deleted filter in ES so we do not want to index them.

invenio_records_resources/services/records/service.py
ppanero: It's a legacy from ES6 IIRC. It was made "optional" in ES7. See [here](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/removal-of-types.html)

kpsherva (approved):

PR #424, opened 2022-11-14
file service: make status aware of fetch files

ppanero:
Sample response with new status:
```
{'mimetype': 'application/zip', 'status': 'pending', 'uri': 'https://zenodo.org/record/4008297/files/dataset.zip#3', ...}
```

### DIFF
--- a/invenio_records_resources/services/files/tasks.py
+++ b/invenio_records_resources/services/files/tasks.py
@@ -22,7 +22,8 @@ def fetch_file(service_id, record_id, file_key):
     file_record = service.read_file_metadata(system_identity, record_id, file_key)
     source_url = file_record.data["uri"]
     # download file
-    with requests.get(source_url, stream=True) as response:
+    # verify=True for self signed certificates by default
+    with requests.get(source_url, stream=True, allow_redirects=True) as response:
         # save file
         service.set_file_content(
             system_identity,
--- a/invenio_records_resources/services/files/schema.py
+++ b/invenio_records_resources/services/files/schema.py
@@ -25,6 +25,8 @@ from marshmallow import (
 from marshmallow.fields import UUID, Dict, Integer, Str
 from marshmallow_utils.fields import GenMethod, Links, SanitizedUnicode, TZDateTime
 
+from .transfer import TransferType
+
 
 class InitFileSchema(Schema):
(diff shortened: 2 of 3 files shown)


zzacharo (approved):
@slint might want to double check this!

invenio_records_resources/services/files/tasks.py
```diff
@@ -22,7 +22,8 @@ def fetch_file(service_id, record_id, file_key):
     file_record = service.read_file_metadata(system_identity, record_id, file_key)
     source_url = file_record.data["uri"]
     # download file
-    with requests.get(source_url, stream=True) as response:
+    # verify=True for self signed certificates by default
+    with requests.get(source_url, stream=True, allow_redirects=True) as response:
```
zzacharo, line 26: why do we need the `allow_redirects` here? what was the problem?

slint: It's more of making sure we don't have issues in case a remote file fetch URL relies on redirects (e.g. imagine CERNBox link redirecting to the new version). It doesn't affect URLs that directly serve the file content, but makes sure we don't store the 3xx response as the file content.

invenio_records_resources/services/files/tasks.py
```diff
@@ -22,7 +22,8 @@ def fetch_file(service_id, record_id, file_key):
     file_record = service.read_file_metadata(system_identity, record_id, file_key)
     source_url = file_record.data["uri"]
     # download file
-    with requests.get(source_url, stream=True) as response:
+    # verify=True for self signed certificates by default
```
slint, line 25: minor: I'm not sure I understand this one. If `verify=True`, that means that certificates (including self-signed) will be attempted to be verified, which might fail. I'm not suggesting to add `verify=False`, but maybe we can add a TODO (or open an issue) to make all these extra parameters configurable via e.g. `RECORDS_RESOURCES_FILES_FETCH_REQUESTS_ARGS = { ... }`.

slint (commented):
LGTM, one small clarification on the comment and potential follow-up

invenio_records_resources/services/files/tasks.py
ppanero: Yes, by default self signed will fail. I can open an issue to make them configurable.

PR #426, opened 2022-11-15
Implemented an all-in-one-download

wgresshoff:
Addresses https://github.com/inveniosoftware/invenio-app-rdm/issues/2037.

:heart: Thank you for your contribution!

### Description

Implemented an all-in-one-download of all files of a record (a streaming solution, so memory and hard-disk consumption is only minimal)

### Checklist

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [x] I've added relevant test cases.
- [ ] I've added relevant documentation.
- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/best-practices/i18n/) (for relevant code).
- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines (for relevant code).
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines (for relevant code).
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines (for relevant code).
- [ ] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [ ] I've NOT included third-party code (copy/pasted source code or new dependencies).

**Third-party code**

If you've added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect](https://github.com/orgs/inveniosoftware/teams/architects/members).

**Reminder**

### DIFF
--- a/invenio_records_resources/resources/files/resource.py
+++ b/invenio_records_resources/resources/files/resource.py
@@ -9,8 +9,10 @@
 
 """Invenio Record File Resources."""
 
+from contextlib import ExitStack
+
 import marshmallow as ma
-from flask import abort, g
+from flask import Response, abort, g, stream_with_context
 from flask_resources import (
     JSONDeserializer,
     RequestBodyParser,
@@ -21,6 +23,7 @@ from flask_resources import (
     response_handler,
     route,
 )
+from zipstream import ZIP_STORED, ZipStream
 
 from ..errors import ErrorHandlersMixin
 from .parser import RequestStreamParser
@@ -64,6 +67,10 @@ class FileResource(ErrorHandlersMixin, Resource):
             route("GET", routes["item"], self.read),
             route("GET", routes["item-content"], self.read_content),
         ]
+        if self.config.allow_archive_download:
+            url_rules += [
+                route("GET", routes["list-archive"], self.read_archive),
+            ]
         if self.config.allow_upload:
             url_rules += [
                 route("POST", routes["list"], self.create),
@@ -174,6 +181,30 @@ class FileResource(ErrorHandlersMixin, Resource):
             abort(404)
 
         return item.send_file(), 200
+
+    @request_view_args
+    def read_archive(self):
+        """Read a zipped version of all files."""
+        id_ = resource_requestctx.view_args["pid_value"]
+        files = self.service.list_files(g.identity, id_)
+
+        def _gen_zipstream():
+            """Generator for the streaming of the zipped file."""
+            zs = ZipStream(compress_type=ZIP_STORED)
+            with ExitStack() as stack:
+                for file_obj in files._results:
+                    fp = stack.enter_context(file_obj.open_stream("rb"))
+                    zs.add(fp, file_obj.key)
+                yield from zs.all_files()
+                yield from zs.finalize()
+
+        return Response(
+            stream_with_context(_gen_zipstream()),
+            mimetype="application/zip",
+            headers={
+                "Content-Disposition": f"attachment; filename={id_}.zip",
+            },
+        )
 
     @request_view_args
     @request_stream
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,5 @@
 ..
-    Copyright (C) 2020 CERN.
+    Copyright (C) 2020-2023 CERN.
 
     Invenio-Records-Resources is free software; you can redistribute it and/or
     modify it under the terms of the MIT License; see LICENSE file for more
@@ -7,6 +7,10 @@
 
 Changes
 =======
+
+Version 1.0.6 (2023-01-23)
+
+- resources: add archive download endpoint for record files
 
 Version 1.0.5 (2023-01-10)
 
--- a/invenio_records_resources/__init__.py
+++ b/invenio_records_resources/__init__.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2020-2022 CERN.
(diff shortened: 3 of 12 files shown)


invenio_records_resources/resources/files/resource.py
```diff
@@ -174,6 +176,32 @@ def read_content(self):
...
+            id_,
+        )
+
+        if files is None:
+            abort(404)
+
+        zs = ZipStream(compress_type=ZIP_STORED)
+        for entry in files.to_dict()["entries"]:
+            file = self.service.get_file_content(g.identity, id_, entry["key"])
+            zs.add(file.get_stream("rb"), entry["key"])
```
ppanero, line 194: major: `get_stream` does not close the fp. See [comment](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/records/api.py#L88). Therefore, this would result in a mem leak. Either use `file.open_stream` (seem to recall there was a reason why it wasn't possible) or close after adding?

wgresshoff: yes, I thought so, too. When I didn't find a solution I analysed what happens and didn't find a clue the files really are left open.

wgresshoff: Using a context manager (like file.open_stream does) leads to the problem that the files are closed to early and they can't be streamed to the zip.

wgresshoff: Ok, I was rechecking and you're right: the files are open! So there is a mem leak and I don't know how to get rid of it.

ppanero: The main issue comes from the fact that the files are streamed to the zip while the response is sent right? In that case, if we need them open till the response is fully sent... I guess we need either a task/callback/descriptor method to close them, not fully sure how to implement it tho... Maybe Alex or someone else has better insights :/

slint (commented):
Just some comments/suggestions on the tests.

kpsherva:
Sorry to come late to the party: In general, it looks good!
I just have one request, since we are implementing it for the whole framework, could you add a feature flag, which allows to ENABLE/DISABLE the feature based on individual needs?

wgresshoff:
That sounds like a good idea! If the all-in-on-download is disabled the response might simply be a 404 (Not found).

invenio_records_resources/resources/files/resource.py
```diff
@@ -174,6 +206,35 @@ def read_content(self):
 
         return item.send_file(), 200
 
+    @request_view_args
+    def read_archive(self):
+        """Read a zipped version of all files."""
+        all_in_one_enabled = current_app.config.get("ALL_IN_ONE_DOWNLOAD_ENABLED", True)
+        if not all_in_one_enabled:
+            abort(404)
```
slint, line 214: minor: instead of checking here, we could conditionally add the endpoint similarly as done with the `allow_upload` resource config variable (see my comment above in the routes).

invenio_records_resources/resources/files/resource.py
```diff
@@ -62,6 +66,7 @@ def create_url_rules(self):
             route("GET", routes["list"], self.search),
             route("GET", routes["item"], self.read),
             route("GET", routes["item-content"], self.read_content),
+            route("GET", routes["item-archive"], self.read_archive),
```
slint, line 69:
```suggestion
        if self.config.allow_archive_download:
            url_rules += [
                route("GET", routes["item-archive"], self.read_archive),
            ]
```

This is the same as the `self.config.allow_upload` below. That way the resource/view functioanlity stays a bit more simple and the enabling/disabling happens on a higher-level.

invenio_records_resources/resources/files/resource.py
```diff
@@ -174,6 +206,35 @@ def read_content(self):
...
+        )
+
+        if files is None:
+            abort(404)
+
+        entries = files.to_dict()["entries"]
+        contents = {}
+        for entry in entries:
+            file = self.service.get_file_content(g.identity, id_, entry["key"])
+            contents[entry["key"]] = file
```
slint, line 229: minor: seeing this part now, it feels like there's a mix of service/presentation layers (see [data flow architecture docs](https://inveniordm.docs.cern.ch/develop/architecture/software/)). I could imagine having all the service calls for listing and opening the files inside one "bigger" service call on the `FileService` (also for efficiency).

slint (commented):
Thanks a lot @wgresshoff for addressing the previous comments! Last comments are about the feature flag implementation, naming consistency, and small leftovers.

If you want we can have a quick chat on Discord as well, just to provide some context before merging.

invenio_records_resources/resources/files/resource.py
wgresshoff: I obviously didn't understand what to do. This is the FileResourceConfig that doesn't contain the setting. And here I can't access `current_app`. Please add some more explanation!

PR #437, opened 2023-01-23
utils: add search query parameters parser

anikachurilova:
* closes https://github.com/inveniosoftware/invenio-banners/issues/11

### DIFF
--- a/invenio_records_resources/services/base/utils.py
+++ b/invenio_records_resources/services/base/utils.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN.
+#
+# Invenio-Records-Resources is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Utils."""
+
+
+def get_search_params(config, params):
+    """Distinguish search parameters."""
+    page = params.get("page", 1)
+    size = params.get(
+        "size",
+        config.search.pagination_options.get("default_results_per_page"),
+    )
+
+    _search_cls = config.search
+
+    _sort_name = (
+        params.get("sort")
+        if params.get("sort") in _search_cls.sort_options
+        else _search_cls.sort_default
+    )
+    _sort_direction_name = (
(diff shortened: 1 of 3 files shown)


Pineirin (approved):

kpsherva (approved):

invenio_records_resources/services/base/utils.py
```diff
@@ -0,0 +1,56 @@
...
+
+    sort = _config.sort_options.get(_sort_name)
+    sort_direction = _config.sort_direction_options.get(_sort_direction_name)
+
+    query_params = params.get("q", "")
+
+    return {
+        "page": page,
+        "size": size,
+        "sort": sort.get("fields"),
```
ntarocco, line 53: it looks like that it will return a `list` of fields when sort is passed, but a `string` when default.

invenio_records_resources/services/base/utils.py
```diff
@@ -0,0 +1,56 @@
...
+# Copyright (C) 2023 CERN.
+#
+# Invenio-Records-Resources is free software; you can redistribute it and/or
+# modify it under the terms of the MIT License; see LICENSE file for more
+# details.
+
+"""Services utils."""
+
+
+def map_search_params(service_search_config, params):
```
ntarocco, line 12: Renamed + added docstring

ntarocco (commented):
It looks like this new func requires a test, there are many `if` and `.get(` :) It should be easy to unit test it.

ntarocco:
I have also removed the export, not very useful for this util

invenio_records_resources/services/base/utils.py
anikachurilova: They are both a `list`: with `?sort=message` param it looks like this `{'page': 1, 'size': 25, 'sort': ['message'], 'sort_direction': <function asc at 0x104e42c10>, 'q': ''}` and without any param like this: `{'page': 1, 'size': 25, 'sort': ['created'], 'sort_direction': <function asc at 0x104e42c10>, 'q': ''}`

ntarocco (approved):

PR #439, opened 2023-02-03
Add file service component for download events

max-moser:
This PR adds a **files service** component that creates and emits a `file-downloaded` stats event whenever the service's `get_file_content(...)` method is called.

I didn't go for a `TaskOp` registered in the currently active unit of work because right now, units of work aren't created for read operations.

WDYT?


invenio_records_resources/resources/records/resource.py
```diff
@@ -106,6 +108,12 @@ def read(self):
             resource_requestctx.view_args["pid_value"],
             expand=resource_requestctx.args.get("expand", False),
         )
+
+        # emit record view stats event
+        emitter = build_event_emitter("record-view")
+        if item is not None and emitter is not None:
+            emitter(current_app, record=item._record, via_api=True)
```
max-moser, line 115: we _could_ emit `record-view` events on `read_draft` (in `invenio-drafts-resources`) as well, but i'm not sure if that's desired?

invenio_records_resources/resources/files/resource.py
```diff
@@ -180,6 +182,12 @@ def read_content(self):
         if item is None:
             abort(404)
 
+        # emit file download stats event
+        obj = item._file.object_version
+        emitter = build_event_emitter("file-download")
```
slint, line 187: minor: since this is re-usable, can it be initialized only once e.g. as a cached property, or another "one-time" init/setup place?

invenio_records_resources/resources/records/resource.py
slint: Agree, it wouldn't make sense, since we also don't have a decent place to expose draft statistics. And admins can probably check access logs to investigate any problems.

slint (approved):
LGTM! One minor comment about the emitter placement/init.

invenio_records_resources/resources/files/resource.py
max-moser: sure – the built `EventEmitter` for a specific event name can be re-used all over the place
given that these emitters are used both here as well as app-rdm, should we maybe put the caching logic into invenio-stats itself?
e.g. on the extension object, something along the lines of `ext.event_emitters[event_name]`?

invenio_records_resources/resources/files/resource.py
```diff
@@ -180,6 +182,12 @@ def read_content(self):
         if item is None:
             abort(404)
 
+        # emit file download stats event
+        obj = item._file.object_version
+        emitter = current_stats.get_event_emitter("file-download")
+        if obj is not None and emitter is not None:
+            emitter(current_app, record=item._record, obj=obj, via_api=True)
```
ppanero, line 189: minor: sending an underscored variable seems to defy the purpose of having it encapsulated. Also `obj` is a bit heavy to be sent? I'm wondering if we can send e.g. files.to_dict() and some sort of serialization of `obj`. 

I seem to recall payloads were not to be discussed now? if so, just ignore my comment ❌  😂

ppanero (approved):

invenio_records_resources/resources/files/resource.py
max-moser: bit of context:
the arguments to `emitter(...)` aren't sent directly as part of a message to the MQ –  rather, they are passed to the registered **event builders** that use them to create a dictionary populated with data fetched from the passed args/kwargs.
after all registered event builders have added their parts, the constructed **event** dictionary is sent off via the MQ.
in the case of the `file-download` events, this event builder will only extract a few properties from the `obj`: https://github.com/inveniosoftware/invenio-app-rdm/pull/2080/files#diff-b4d96b23efeba4170d5d72d1330d7f17d19ebfb5e07d211939bfd720543dbb66R23-R50
the default event builders for InvenioRDM can be found in the app-rdm config: https://github.com/inveniosoftware/invenio-app-rdm/pull/2080/files#diff-988eb51a0bc0a8bf6efd0b04d1bb6b6d8f24cc80504ea36699a202904a20af4dR876-R879

**but yes**, good point!
we could probably send some version of `record.to_dict()` and `obj.to_dict()` to the event builders rather than the actual objects themselves... that would probably help out a lot in the future when instead of building & emitting events synchronously like here, we'd add a (probably MQ-based) event bus :smile:

invenio_records_resources/resources/files/resource.py
slint: Oof, forgot to reply, but yeah `invenio-stats` would be the place, and the current implementation looks good :)

invenio_records_resources/resources/files/resource.py
ppanero: That makes, sense. In my head it would make more sense not to access an `_` value. However, more sense would make an event bus. So for me is fine if left like this since it will be re-worked (hopefully) at some point. So up to you. Thanks a lot for the explanation!

setup.cfg
```diff
@@ -49,6 +50,7 @@ tests =
     pytest-black>=0.3.0
     invenio-app>=1.3.2
     invenio-db[postgresql,mysql,versioning]>=1.0.14,<2.0.0
+    invenio-stats>=2.0.0,<3.0.0
```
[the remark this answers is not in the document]
kpsherva:
```suggestion
    invenio-stats>=3.0.0,<4.0.0
```
because we need to take over the v2.0.0 slot for other changes

jrcastro2 (approved):

invenio_records_resources/resources/records/resource.py
```diff
@@ -106,6 +108,12 @@ def read(self):
             resource_requestctx.view_args["pid_value"],
             expand=resource_requestctx.args.get("expand", False),
         )
+
+        # emit record view stats event
+        emitter = current_stats.get_event_emitter("record-view")
+        if item is not None and emitter is not None:
+            emitter(current_app, record=item._record, via_api=True)
```
ntarocco, line 115: I guess this was already discussed, but I would have instead:
1. move the emitting part inside the `read` service method
2. pass to the service an extra kwargs to give context to that call, e.g. `is_api=[True|False]`, with a default value or a mandatory value (to avoid forgetting to pass it).
 
This would allow code duplication when adding this in the Flask `view`.

ntarocco (commented):
IMO, this requires a quick chat. Ping @kpsherva and @slint too.

invenio_records_resources/resources/records/resource.py
max-moser: Summary of the brief discussion just now:
The events fetch some request-dependent information (e.g. client IP address) that isn't necessarily available to the service right now. This information would have to be passed in some reliable way.
Further, it might be that the service layer is called from other places that we do not want to count, e.g. as part of a CLI command (like we do in `Invenio-Utilities-TUW` sometimes).
Basically, it was found that the resources & view functions are an okay place, because those are the calls we're interested in counting.

kpsherva: please add a comment in the code as well :)

ntarocco: > Summary of the brief discussion just now: The events fetch some request-dependent information (e.g. client IP address) that isn't necessarily available to the service right now. This information would have to be passed in some reliable way. 

On this, not sure what is the issue around "reliable way". In any case, see below.

> Further, it might be that the service layer is called from other places that we do not want to count, e.g. as part of a CLI command (like we do in `Invenio-Utilities-TUW` sometimes).

Agree very much on this and, for example, the fact that a service returns a `result_item`, instead of the record itself, is an issue when calling services from services.
It feels like something we are not doing 100% correctly. Let's go ahead with this to avoid blocking this PR further, we can always refactor later on.

PR #441, opened 2023-02-10
service: add Record Indexer Service mixin

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1163

### DIFF
--- a/invenio_records_resources/services/records/service.py
+++ b/invenio_records_resources/services/records/service.py
@@ -27,12 +27,13 @@ from ..uow import RecordCommitOp, RecordDeleteOp, unit_of_work
 from .schema import ServiceSchemaWrapper
 
 
-class RecordService(Service):
-    """Record Service."""
+class RecordIndexerMixin:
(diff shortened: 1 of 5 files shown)


ppanero (approved):
once the test pass

invenio_records_resources/services/records/service.py
```diff
@@ -27,12 +27,16 @@
...
-    """Record Service."""
+class RecordIndexerMixin:
+    """Mixin class to define record indexer."""
+
+    def __init__(self, config):
+        """Constructor.
+
+        :param config: A service configuration
+        """
+        self.config = config
```
slint, line 38: moderate: shouldn't this stay in the `RecordService`, or even be moved up to `Service`?

jrcastro2: It's already in the `Service`, I added it there as well as it needs the configuration to work, ideally it should not be used alone, but if done it would fail if the config is missing.

Issue #438, opened 2023-02-02
files: serialize `uri` only for non-local files


slint:
In https://github.com/inveniosoftware/invenio-records-resources/blob/872683b32b4b7998acce4afdf74a64f0a52261fd/invenio_records_resources/services/files/schema.py#L88 we're always including in serialization the file's URI.

This is problematic, especially for local files (i.e. `FileInstance.storage_class: "L"`), since it's exposing internals of file storage (local filesystem storage paths, etc.). We should:

- [ ] Exclude the `uri` for `LOCAL (L)` storage class files
- [ ] Evaluate for which cases we actually want to serialize this attribute at all:
  - [ ] The `REMOTE (R)` storage class is meant for linking remote/external files that are not stored on the instance, but we want to be part of the record (e.g. because of licensing issues). When requesting `GET /api/records/{id}/files/{filename}/content`, we get a redirect response to the `uri`. In a way the `uri` can become public anyway (in the redirect response), so exposing it doesn't impose any danger. 
  - [ ] The `FETCH (F)` storage class is used to mark that this file will be fetched by a background task which on success will convert it into a `LOCAL (L)` file and mark the file as "committed" (so that the record can be published). This means that it will never be the case that a publicly published record would have a file with the `FETCH (F)` storage class, and thus there's no risk of exposing a "private" `uri` to the public (and the uploader already knows the fetch `uri`, so no danger there either). In that sense, it's probably ok to serialize `uri` for this storage class

PR #443, opened 2023-02-13
files : serialize uri only for non-local files

TLGINO:
* closes https://github.com/inveniosoftware/invenio-records-resources/issues/438

### DIFF
--- a/invenio_records_resources/services/files/schema.py
+++ b/invenio_records_resources/services/files/schema.py
@@ -85,7 +85,11 @@ class InitFileSchema(Schema):
         if data.file:
             # mandatory fields
             data["storage_class"] = data.file.storage_class
-            data["uri"] = data.file.uri  # TODO: only serialize for external?
+            data["uri"] = data.file.uri
+
+            # If Local -> remove uri as it contains internal file storage info
+            if not TransferType(data["storage_class"]).is_serializable():
+                data.pop("uri")
 
             # optional fields
             fields = ["checksum", "size"]
--- a/invenio_records_resources/services/files/transfer.py
+++ b/invenio_records_resources/services/files/transfer.py
@@ -45,6 +45,10 @@ class TransferType(str, Enum):
         """Return if the type represents a completed transfer."""
         return self in [TransferType.LOCAL, TransferType.REMOTE]
 
+    def is_serializable(self):
+        """Return if the type represents a localy available file."""
+        return self != TransferType.LOCAL
+
 
 class BaseTransfer(ABC):
     """Local transfer."""
--- a/tests/services/files/test_file_service.py
+++ b/tests/services/files/test_file_service.py
@@ -99,11 +99,13 @@ def test_file_flow(file_service, location, example_file_record, identity_simple)
     result = file_service.list_files(identity_simple, recid)
     assert result.to_dict()["entries"][0]["key"] == file_to_initialise[0]["key"]
     assert result.to_dict()["entries"][0]["storage_class"] == "L"
+    assert "uri" not in result.to_dict()["entries"][0]
 
     # Read file metadata
     result = file_service.read_file_metadata(identity_simple, recid, "article.txt")
     assert result.to_dict()["key"] == file_to_initialise[0]["key"]
     assert result.to_dict()["storage_class"] == "L"
+    assert "uri" not in result.to_dict()
 
     # Retrieve file
     result = file_service.get_file_content(identity_simple, recid, "article.txt")
@@ -204,6 +206,7 @@ def test_external_file_simple_flow(
     result = result.to_dict()
     assert result["key"] == file_to_initialise[0]["key"]
     assert result["storage_class"] == "L"  # changed after commit
+    assert "uri" not in result
 
     # Retrieve file
     result = file_service.get_file_content(identity_simple, recid, "article.txt")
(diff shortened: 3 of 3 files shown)


invenio_records_resources/services/files/schema.py
```diff
@@ -85,7 +85,20 @@ def fields_from_file_obj(self, data, **kwargs):
         if data.file:
             # mandatory fields
             data["storage_class"] = data.file.storage_class
-            data["uri"] = data.file.uri  # TODO: only serialize for external?
+            data["uri"] = data.file.uri
+
+            # If Local -> remove uri as it contains internal file storage info
+            if data["storage_class"] == "L":
```
ntarocco, line 91: What about using the "enum" instead?
```suggestion
            if data["storage_class"] == TransferType.LOCAL:
```

invenio_records_resources/services/files/schema.py
```diff
@@ -85,7 +85,20 @@ def fields_from_file_obj(self, data, **kwargs):
...
+                data.pop("uri")
+
+            # [TODO] Remote and Fetch
```
ntarocco, line 94: Why this TODO? Is it because you are not sure what should be the wanted behaviour here?
If that's the case, it looks like that the answers are in the issue ;)

invenio_records_resources/services/files/schema.py
```diff
@@ -85,7 +85,20 @@ def fields_from_file_obj(self, data, **kwargs):
...
+            elif data["storage_class"] == "F":
+                pass
+            elif data["storage_class"] == "R":
+                pass
+            else:
+                # Error?
```
ntarocco, line 100: I would not raise an error, I would `.pop`. by default? Better safe that sorry?

invenio_records_resources/services/files/schema.py
TLGINO: I put it as a TODO "discussion"
But yes for both of the cases (Fetch and Remote) there should be no need to modify the URI
I'll remove them

invenio_records_resources/services/files/schema.py
TLGINO: You're right I'll change thx!

invenio_records_resources/services/files/schema.py
```diff
@@ -85,7 +85,18 @@ def fields_from_file_obj(self, data, **kwargs):
         if data.file:
             # mandatory fields
             data["storage_class"] = data.file.storage_class
-            data["uri"] = data.file.uri  # TODO: only serialize for external?
+            data["uri"] = data.file.uri
+
+            # If Local -> remove uri as it contains internal file storage info
+            if data["storage_class"] == TransferType.LOCAL:
```
ppanero, line 91: Could we use the `Transfer.` class factory with the `is_completed` function? That way we avoid exposing/coupling the TransferType categories.

invenio_records_resources/services/files/schema.py
ppanero: If we do not raise an error the record will get published with incomplete files:
1. I add a FETCH file
2. I click publish

Once the record has been published the files cannot be added (bucket is locked). This is a limitation from the implementation, but records cannot be published until their files have been downloaded.

invenio_records_resources/services/files/schema.py
ppanero, line 91: Can we add a method in the Transfer class, for example "local_uri" (I dont like this name). To avoid coupling the TransferTypes (data layer) to the service serialization?

ntarocco: Maybe @TLGINO understood, but I am not sure I did fully :) It sounds like a good idea, could you elaborate what you mean? :) Thank you!

ppanero: Sorry, I should have explained better. In invenio-records-resources the [TransferType] class has a method to specify if the transfer has finished or not. I was wondering if we could make a similar method to enable/disable serialization (e.g. is_serializable). That way we would abstract this behavior (i.e. in case we create more transfertype we do not need to modify the serialization/service schema). It would also simplify the code:

```python
if not TransferType(data["storage_class"]).is_serializable:
                 data.pop("uri")
```

invenio_records_resources/services/files/schema.py
```diff
@@ -85,7 +85,11 @@ def fields_from_file_obj(self, data, **kwargs):
         if data.file:
             # mandatory fields
             data["storage_class"] = data.file.storage_class
-            data["uri"] = data.file.uri  # TODO: only serialize for external?
+            data["uri"] = data.file.uri
+
+            # If Local -> remove uri as it contains internal file storage info
+            if data["storage_class"].is_serializable():
```
ppanero, line 91: sorry, should this not be the opposite? if it is serializable that means we *can* put it in the payload, if it is not we remove it?

it really depends on the meaning of the function, WDYT?

TLGINO: woops yes lol, will correct

TLGINO: nvm all good!

ppanero (commented):
I'm very sorry for all the back and forth, I didnt notice this before. Could we add some tests to validate this?
e.g. 
- check there IS uri in the response here https://github.com/inveniosoftware/invenio-records-resources/blob/master/tests/services/files/test_file_service.py#L194
- then here there is NO uri https://github.com/inveniosoftware/invenio-records-resources/blob/master/tests/services/files/test_file_service.py#L206

ppanero (approved):

PR #444, opened 2023-02-17
services: enable count on searches

jrcastro2:
* param version is not allowed in count queries, therfore has to be optional
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2092


ppanero:
This PR is an utility and can be merged independently of its issue. However, it needs work to be merged (i.e. pass tests)

ppanero:
Tests were failing on BLACK, I've pushed a commit to fix it

ppanero (approved):

PR #449, opened 2023-03-01
Fix "size" field missing on empty files

max-moser:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2032

This PR addresses the issue that empty files (with size 0) blow up record landing pages with an error message stating `jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'size'`.

I have a feeling that there's a better way of doing this that involves the [`schema.fields_from_file_obj`](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/files/schema.py#L77), but I'll have to check some more.
Potentially it also has to do with 0 being the "null" value for integers in the database model?

### DIFF
--- a/tests/resources/test_files_resource.py
+++ b/tests/resources/test_files_resource.py
@@ -163,6 +163,52 @@ def test_files_api_flow(client, search_clear, headers, input_data, location):
     assert len(res.json["entries"]) == 0
 
 
+def test_empty_file(client, search_clear, headers, input_data, location):
+    """Test if an empty file works properly."""
+    # Initialize a draft
+    res = client.post("/mocks", headers=headers, json=input_data)
+    assert res.status_code == 201
+    id_ = res.json["id"]
+    assert res.json["links"]["files"].endswith(f"/api/mocks/{id_}/files")
+
+    # Initialize files upload
+    res = client.post(
+        f"/mocks/{id_}/files",
+        headers=headers,
+        json=[
+            {"key": "empty", "title": "Zero-length test file"},
+        ],
+    )
+    assert res.status_code == 201
+
+    # Upload the empty file
+    res = client.put(
+        f"/mocks/{id_}/files/empty/content",
+        headers={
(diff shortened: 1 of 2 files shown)


max-moser:
Looking just a little bit further into the `InitFileSchema` revealed a bug while dumping optional fields, which is now being fixed with this PR.

slint (approved):

slint:
I would say maybe we add a test case as well, since 0-byte sized files aren't that uncommon... 

But then what I have in mind are more integration-style testing, i.e. covering a set of common files use-cases ("normal" file, exotic unicode filenames, 0-byte sized, etc.) and how they behave throughout common actions in the system (upload, download, get metadata, etc.). And I'm not sure if this style of tests is for this module, or better higher up in the hierarchy (e.g. `rdm-records` or `app-rdm`).

ppanero (approved):
I agree with @slint. I would actually add those tests in this module since it is where the functionality is added. Otherwise we might end up detecting errors higher up (e.g. app-rdm/rdm-records) and having to make changes at this level.

tests/resources/test_files_resource.py
```diff
@@ -163,6 +163,52 @@ def test_files_api_flow(client, search_clear, headers, input_data, location):
...
+    # Commit the uploaded file
+    res = client.post(f"/mocks/{id_}/files/empty/commit", headers=headers)
+    assert res.status_code == 200
+
+    # Get the file metadata, check if everything (especially size) is present
+    res = client.get(f"/mocks/{id_}/files/empty", headers=headers)
+    assert res.status_code == 200
+    assert res.json["key"] == "empty"
+    assert res.json["checksum"] == "md5:d41d8cd98f00b204e9800998ecf8427e"
+    assert res.json["size"] == 0
```
max-moser, line 204: without the fix, this line here will fail with a `KeyError: 'size'`

slint (approved):

PR #452, opened 2023-03-03
resolve deleted resources

zzacharo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2063

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,10 @@
 Changes
 =======
 
+Version 1.3.1 (2023-03-23)
+
+- custom-fields: control `field_cls` customization
+
 Version 1.3.0 (2023-03-17)
 
 - errors: add FileKeyNotFoundError
--- a/invenio_records_resources/__init__.py
+++ b/invenio_records_resources/__init__.py
@@ -10,6 +10,6 @@
 
 from .ext import InvenioRecordsResources
 
-__version__ = "1.3.0"
+__version__ = "1.3.1"
 
 __all__ = ("__version__", "InvenioRecordsResources")


invenio_records_resources/services/records/results.py
```diff
@@ -352,13 +353,30 @@ def _find_fields(self, service, value):
...
+                # possible deletions
+                found_values.add(value)
                 for field in self._find_fields(service, value):
                     field.add_dereferenced_record(service, value, hit)
 
+            not_found_values = values - found_values
+            if not_found_values:
+                for value in not_found_values:
+                    try:
+                        service.read(identity, value)
```
zzacharo, line 373: maybe we could avoid that by either
1. Assuming that the record is deleted if not found
2. Change resource status to "not_found" rather than "deleted"

invenio_records_resources/services/records/results.py
```diff
@@ -352,13 +353,30 @@ def _find_fields(self, service, value):
...
+                    except PIDDeletedError:
+                        for field in self._find_fields(service, value):
+                            field.add_dereferenced_record(
+                                service, value, {"id": value, "status": "deleted"}
```
zzacharo, line 377: Is it worth to move the default value to `ExpandableField` as new property that is defined by the user? I.e moving the delegation to when declaring the expandable fields rather than when the `Resolver` class when picking fields?

invenio_records_resources/services/records/results.py
lnielsen: is worth trying to do the ``read()``...... couldn't this potentially become quite expensive?

invenio_records_resources/services/records/results.py
```diff
@@ -264,6 +264,14 @@ def get_value_service(self, value):
         """Return the value and the service to fetch the referenced record."""
         return None, None
 
+    @abstractmethod
+    def tombstone(self, value):
+        """Return the tombstone representation of the value.
+
+        This is used to be returned when a value cannot be resolved
+        """
+        return {}
```
ppanero, line 273: derivate from the file comment: If it is an `abstractmethod` the inheriting class is forced to reimplement so it does not make sense to have a return value. Alternatively, if we want to have a default (e.g. `{}`) then the decorator should be removed.
```suggestion
     pass
```

Example:

```python
>>> class A(ABC):
...     @abstractmethod
...     def test(self):
...             return {}
...
>>> class B(A):
...     pass
...
>>> B().test()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Can't instantiate abstract class B with abstract method test
```

invenio_records_resources/services/records/results.py
```diff
@@ -352,12 +362,26 @@ def _find_fields(self, service, value):
 
     def _fetch_referenced(self, grouped_values, identity):
         """Search and fetch referenced recs by ids."""
+
+        def _add_dereferenced_record(service, value, resolved_rec):
+            """Helper function to set the dereferenced record to the service."""
+            for field in self._find_fields(service, value):
+                field.add_dereferenced_record(service, value, resolved_rec)
+
         for service, values in grouped_values.items():
```
ppanero, line 371: There is a lot of nested looping in this function and it is hard to understand, maybe adding more comments would help.

invenio_records_resources/services/records/results.py
ppanero, line 1: major: Given the tests and the usage of `ExpandableField`, I think it should not be an `ABC`:
- It has default behaviour for many `abstractmethod`, which by definition should not
- The `abstractmethod` forces doing `super()` calls to perform the default behaviour. Without the decorator, all the changes done in `tests/services/test_results_expand.py` could be removed since the parent function would be called by default.

tests/services/test_results_expand.py
```diff
@@ -45,6 +45,10 @@ def read_many(self, identity, ids):
 
 
 class CreatedByExpandableField(ExpandableField):
+    def tombstone(self, value):
```
ppanero, line 48: remove all implementations of `tombstone` when the parent class is not ABC anymore

invenio_records_resources/services/records/results.py
```diff
@@ -264,6 +264,14 @@ def get_value_service(self, value):
         """Return the value and the service to fetch the referenced record."""
         return None, None
 
+    @abstractmethod
+    def tombstone(self, value):
```
ppanero, line 268: The datamodel is known by the service layer, would it make sense to have something like `service.tombstone(value)`? There would not be need for perm checks or anything, just to encapsulate correctly in the service and make it "datamodel" aware (see more in the communities PR, about hardcoding the dict).

ppanero (approved):

invenio_records_resources/services/records/results.py
```diff
@@ -279,6 +292,10 @@ def add_service_value(self, service, value):
 
     def add_dereferenced_record(self, service, value, resolved_rec):
         """Save the dereferenced record."""
+        if resolved_rec is None:
+            resolved_rec = self.defaults({"id": value})
+            # mark the record as a "ghost" record
+            resolved_rec["is_ghost"] = True
```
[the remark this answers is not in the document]
zzacharo: I left it here intentionally, so it's kind of forced instead of relying on each concrete implementation to add it. That way, we can eventually rely on this key to be there in all different views...

zzacharo:
continues #461

PR #454, opened 2023-03-09
Add new `FileKeyNotFoundError`

max-moser:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1998

This PR introduces a new custom error type: `FileKeyNotFoundError` to denote when the service was trying to access a non-existing file key for a record.

### DIFF
--- a/invenio_records_resources/resources/errors.py
+++ b/invenio_records_resources/resources/errors.py
@@ -33,6 +33,7 @@ from werkzeug.routing import BuildError
 from ..errors import validation_error_to_list_errors
 from ..services.errors import (
     FacetNotFoundError,
+    FileKeyNotFoundError,
     PermissionDeniedError,
     QuerystringValidationError,
     RevisionIdMismatchError,
@@ -156,6 +157,12 @@ class ErrorHandlersMixin:
             lambda e: HTTPJSONException(
                 code=404,
                 description=f"Facet {e.vocabulary_id} not found.",
+            )
+        ),
+        FileKeyNotFoundError: create_error_handler(
+            lambda e: HTTPJSONException(
+                code=404,
+                description=str(e),
             )
         ),
         JSONDecodeError: create_error_handler(
--- a/invenio_records_resources/services/errors.py
+++ b/invenio_records_resources/services/errors.py
@@ -55,3 +55,17 @@ class FacetNotFoundError(Exception):
         """Initialise error."""
         self.vocabulary_id = vocabulary_id
         super().__init__(_("Facet {vocab} not found.").format(vocab=vocabulary_id))
+
+
+class FileKeyNotFoundError(Exception):
+    """Error denoting that a record doesn't have a certain file."""
+
+    def __init__(self, recid, file_key):
+        """Constructor."""
+        super().__init__(
+            _("Record '{recid}' has no file '{file_key}'.").format(
+                recid=recid, file_key=file_key
+            )
+        )
+        self.recid = recid
+        self.file_key = file_key
--- a/invenio_records_resources/services/files/service.py
+++ b/invenio_records_resources/services/files/service.py
@@ -10,6 +10,7 @@
 """File Service API."""
 
 from ..base import LinksTemplate, Service
+from ..errors import FileKeyNotFoundError
 from ..records.schema import ServiceSchemaWrapper
 from ..uow import RecordCommitOp, unit_of_work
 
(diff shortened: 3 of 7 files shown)


invenio_records_resources/services/files/service.py
```diff
@@ -49,11 +50,19 @@ def check_permission(self, identity, action_name, **kwargs):
...
+        If a ``file_key`` is specified and the record in question doesn't have a file
+        for that key, a ``FileKeyNotFoundError`` will be raised.
+        """
         # FIXME: Remove "registered_only=False" since it breaks access to an
         # unpublished record.
         record = self.record_cls.pid.resolve(id_, registered_only=False)
         self.require_permission(identity, action, record=record, file_key=file_key)
+
+        if file_key is not None and file_key not in record.files:
+            raise FileKeyNotFoundError(id_, file_key)
```
max-moser, line 64: moving the check into this helper function simplifies the service methods a bit, by reducing duplication.

invenio_records_resources/services/files/tasks.py
```diff
@@ -19,17 +19,20 @@
 def fetch_file(service_id, record_id, file_key):
     """Fetch file from external storage."""
     service = current_service_registry.get(service_id)
+    # FIXME HERE
```
max-moser, line 22: i'm not really sure how to handle the errors in the tasks... should we put a `try/except` around and log the error, or should we just simply ignore the exception?

invenio_records_resources/resources/errors.py
```diff
@@ -158,6 +159,12 @@ class ErrorHandlersMixin:
                 description=f"Facet {e.vocabulary_id} not found.",
             )
         ),
+        FileKeyNotFoundError: create_error_handler(
+            lambda e: HTTPJSONException(
+                code=404,
+                description=f"The record has no file named '{e.file_key}'.",
```
rekt-hard, line 165: minor: align message/description of exception on service and resource level

invenio_records_resources/services/files/service.py
rekt-hard: question: I am not sure what the permission check will do or access exactly. Would it make sense to ensure that the file exists before checking permissions?

invenio_records_resources/services/files/tasks.py
rekt-hard: I would definitely log the errors, so it can be monitored. A general `try/except` around the task might be enough, imo.

In general and just as an idea: To not duplicate it in every task, how about introducing a general task error handler, or sub-classing the celery task and define an error handler? Then this task could be used almost everywhere?

rekt-hard (commented):
Looking good. Just minor things

invenio_records_resources/services/files/tasks.py
max-moser: I've added the `try/except` with logging in the tasks themselves for now, and I'd create a follow-up issue for the improvement suggestion.
Where do you think it would be most appropriate?

invenio_records_resources/services/errors.py
```diff
@@ -55,3 +55,13 @@ def __init__(self, vocabulary_id):
...
         self.vocabulary_id = vocabulary_id
         super().__init__(_("Facet {vocab} not found.").format(vocab=vocabulary_id))
+
+
+class FileKeyNotFoundError(Exception):
+    """Error denoting that a record doesn't have a certain file."""
+
+    def __init__(self, recid, file_key):
+        """Constructor."""
+        super().__init__(f"Record '{recid}' has no file '{file_key}'.")
```
ppanero, line 65: This is missing formatting and translation?

```suggestion
        super().__init__(_("Record {recid} has no file {file_key}.").format(recid=recid, file_key=file_key))
```

invenio_records_resources/services/files/service.py
```diff
@@ -49,11 +50,19 @@ def check_permission(self, identity, action_name, **kwargs):
...
+
+        If a ``file_key`` is specified and the record in question doesn't have a file
+        for that key, a ``FileKeyNotFoundError`` will be raised.
+        """
         # FIXME: Remove "registered_only=False" since it breaks access to an
         # unpublished record.
         record = self.record_cls.pid.resolve(id_, registered_only=False)
         self.require_permission(identity, action, record=record, file_key=file_key)
+
+        if file_key is not None and file_key not in record.files:
```
ppanero, line 63:
```suggestion
        if not file_key and file_key not in record.files:
```

ppanero (approved):
Just some very minor comments, if rebased and test pass 🚢

Issue #447, opened 2023-02-27
search: defining a query transformer invalidates all other terms search


alejandromumo:
When a query transformer is defined, all the other search terms (not defined in the transformer mapping) are basically "ignored" by the query parser:

https://github.com/inveniosoftware/invenio-records-resources/blob/41d355f470deb0bd76b798e0399033f4d2ed96a0/invenio_records_resources/services/records/queryparser/query.py#L87-L94

https://github.com/inveniosoftware/invenio-records-resources/blob/41d355f470deb0bd76b798e0399033f4d2ed96a0/invenio_records_resources/services/records/queryparser/transformer.py#L36-L40

When the exception is thrown, a `multi match` query is used instead.

Steps to reproduce:
- Enable search query transformer ([docs](https://inveniordm.docs.cern.ch/develop/howtos/search_terms_migration/))
- Add a mapping (e.g `{'title': 'metadata.title'}`
- Run the instance
- Navigate to UI
- Search for a record (e.g. `metadata.title:"<a title>"`)
- Validate that the term was not taken into consideration. Instead, multiple records are displayed (if the search value matches multiple records)

alejandromumo:
For now we will complete the transformer mapping with the missing fields .

alejandromumo:
Some notes about the latest discussion:

- Mapping *every* term in invenio is not ideal for many reasons (time consuming, not maintainble, etc). So we are trying to figure out a "smarter" way of fixing this issue. 
- Initially the `QueryTransformer` feature also had a "security" side: terms that were not defined in the mapping were not allowed. This was particulary useful to prevent some terms of being queried, e.g. for [users](https://github.dev/inveniosoftware/invenio-users-resources/blob/be712725fc2e4a6da43219450564f32572efbfc3/invenio_users_resources/services/users/config.py#L43). This worked as a "safelist" in the client-side (e.g. the search engine is not involved).

I analysed how the query parser works and these are my thoughts:

- Currently, the "safelist" works because we default the query to a `MultiMatch` with specified `fields`. This means that elasticsearch will try to match the values only to the given fields. 
    - Personally, it feels more like a side effect. 
    - Also, if no `fields` are specified, [every field is used in the search](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html#field-boost)
- Personally, I think that having a "safelist" of terms and a "transformer mapping" are two different transformations on the query.
    - We might want none, one of either or both.
[...]

PR #462, opened 2023-03-24
queryparser: added safelisting of terms

alejandromumo:
closes https://github.com/inveniosoftware/invenio-records-resources/issues/447

### DIFF
--- a/invenio_records_resources/services/records/queryparser/transformer.py
+++ b/invenio_records_resources/services/records/queryparser/transformer.py
@@ -24,24 +24,30 @@ from invenio_records_resources.services.errors import QuerystringValidationError
 class SearchFieldTransformer(TreeTransformer):
     """Transform from user-friendly field names to internal field names."""
 
-    def __init__(self, mapping, *args, **kwargs):
+    def __init__(self, mapping, allow_list, *args, **kwargs):
         """Constructor."""
         self._mapping = mapping
+        self._allow_list = allow_list
         super().__init__(self, *args, **kwargs)
 
     @classmethod
-    def factory(cls, mapping=None):
+    def factory(cls, mapping=None, allow_list=None):
         """Create a new field transformer."""
-        return partial(cls, mapping or {})
+        return partial(cls, mapping or {}, allow_list or [])
 
     def visit_search_field(self, node, context):
         """Visit a search field."""
-        if node.name not in self._mapping:
+        # Use the node name if not mapped for transformation.
+        term_name = self._mapping.get(node.name, node.name)
+
+        # If a allow list exists, the term must be allowed to be queried.
+        if self._allow_list and not term_name in self._allow_list:
             raise QuerystringValidationError(
                 _("Invalid search field: {field_name}.").format(field_name=node.name)
             )
-        else:
-            new_node = node.clone_item()
-            new_node.name = self._mapping[node.name]
-            new_node.children = list(self.clone_children(node, new_node, context))
-            yield new_node
+
+        # Returns a copy of the node.
+        new_node = node.clone_item()
+        new_node.name = term_name
+        new_node.children = list(self.clone_children(node, new_node, context))
+        yield new_node
--- a/tests/services/test_service_queryparser.py
+++ b/tests/services/test_service_queryparser.py
@@ -96,21 +96,23 @@ def test_multimatch(parser, query):
 
 
 @pytest.mark.parametrize(
-    "query,transformed_query,query_type",
+    "query,transformed_query,query_type, allow_list",
     [
-        ("title:test", "metadata.title:test", "query_string"),
-        ("title:(test test)", "metadata.title:(test test)", "query_string"),
-        ("title:[1 TO 5]", "metadata.title:[1 TO 5]", "query_string"),
-        # Invalid field names results in multi_match
-        ("description:test", "description:test", "multi_match"),
-        ("metadata.title:test", "metadata.title:test", "multi_match"),
+        ("title:test", "metadata.title:test", "query_string", None),
+        ("title:(test test)", "metadata.title:(test test)", "query_string", None),
+        ("title:[1 TO 5]", "metadata.title:[1 TO 5]", "query_string", None),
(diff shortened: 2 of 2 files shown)


invenio_records_resources/services/records/queryparser/transformer.py
```diff
@@ -24,24 +24,30 @@
 class SearchFieldTransformer(TreeTransformer):
     """Transform from user-friendly field names to internal field names."""
 
-    def __init__(self, mapping, *args, **kwargs):
+    def __init__(self, mapping, safelist, *args, **kwargs):
```
alejandromumo, line 27: This can also be implemented in the `QueryParser` as a "step" of the parsing process. 
I added it here for now since we're visiting one field at the time (using `luqum`). So it was easier to do it here and didn't feel completely wrong.

But please challenge it :)

alejandromumo:
I can fix tests when this is accepted, since now it's a "WIP" I will not fix them and focus on other tasks.

invenio_records_resources/services/records/queryparser/transformer.py
```diff
@@ -24,24 +24,30 @@
...
+        return partial(cls, mapping or {}, safelist or [])
 
     def visit_search_field(self, node, context):
         """Visit a search field."""
-        if node.name not in self._mapping:
+        # Use the node name if not mapped for transformation.
+        term_name = self._mapping.get(node.name, node.name)
+
+        # If a safelist exists, the term must be safelisted to be queried.
+        if self._safelist and not term_name in self._safelist:
```
ppanero, line 44: Do we have a case where we would not want a safelist? records? This is to make our lives easier when defining the parsing?

alejandromumo: Sorry for the late reply!

Yes there are searches that we do not want / need a safelist. Initially the terms mapping also acted as a safelist and that was particularly useful for users becasuse some terms are not meant to be queried.

However, since we shipped two features in one (mapping + safelisting) it also meant that if a transform map was defined, then any other terms are considered were blocked (e.g. can't be queried at all). In fact, this had a side effect on which the transformation map needed things like `{"email": "email"}` since the term was not transformed AND being blocked if not defined.

TL;DR : yes we do have cases where we do not want to have a safelist. And yes, this separation of features makes our lives easier.

ppanero (approved):

ppanero (approved):
I have mostly questions for my understanding, I don't think any should block the merge.

tests/services/test_service_queryparser.py
```diff
@@ -96,21 +96,23 @@ def test_multimatch(parser, query):
...
-        ("description:test", "description:test", "multi_match"),
-        ("metadata.title:test", "metadata.title:test", "multi_match"),
+        ("title:test", "metadata.title:test", "query_string", None),
+        ("title:(test test)", "metadata.title:(test test)", "query_string", None),
+        ("title:[1 TO 5]", "metadata.title:[1 TO 5]", "query_string", None),
+        # Term is allowed and not mapped, results in a query string with the original query
+        ("description:test", "description:test", "query_string", None),
+        ("metadata.title:test", "metadata.title:test", "query_string", None),
+        # Terms might be mapped but not allowed, results in multi_match with the original query
+        ("title:test", "title:test", "multi_match", ["metadata.date"]),
```
ppanero, line 108: I'm missing something (because tests pass). Why isn't this entering `transformer.py#44`? The query term `title` is in the mapping and there is an allowed list without it so:

```
if self._allow_list and not term_name in self._allow_list:

if ["metadata.date"] and not `title` in ["metadata.date"]:

True and not False:
```

I'm puzzled 😂

ppanero: https://github.com/inveniosoftware/invenio-records-resources/blob/0547e5d7cb7d30cca5c470e6fb264de408ce77ed/invenio_records_resources/services/records/queryparser/query.py#L92

ppanero (commented):
Discussed with Nico and Karolina. The two query parser issues will be tackled independenty here https://github.com/inveniosoftware/invenio-records-resources/issues/464.

Therefore, there is only a need for updating (ref [search](https://github.com/search?q=org%3Ainveniosoftware%20QueryParser&type=code))
- https://github.com/inveniosoftware/docs-invenio-rdm/blob/63923ea359ad4ab41bc37440098d59d0fe494d26/docs/develop/howtos/search_terms_migration.md?plain=1#L23
- https://github.com/inveniosoftware/invenio-users-resources/blob/10d739c945e371366026d6f0156566c87cfb5f1f/invenio_users_resources/services/groups/config.py#L38 <- Not sure about this one

ppanero:
Needs updating (ref [search](https://github.com/search?q=org%3Ainveniosoftware%20QueryParser&type=code))
- https://github.com/inveniosoftware/docs-invenio-rdm/blob/63923ea359ad4ab41bc37440098d59d0fe494d26/docs/develop/howtos/search_terms_migration.md?plain=1#L23
- https://github.com/inveniosoftware/invenio-users-resources/blob/10d739c945e371366026d6f0156566c87cfb5f1f/invenio_users_resources/services/groups/config.py#L38 <- Not sure about this one

invenio_records_resources/services/records/queryparser/transformer.py
```diff
@@ -24,24 +24,30 @@
...
                 _("Invalid search field: {field_name}.").format(field_name=node.name)
             )
-        else:
-            new_node = node.clone_item()
-            new_node.name = self._mapping[node.name]
-            new_node.children = list(self.clone_children(node, new_node, context))
-            yield new_node
+
+        # Returns a copy of the node.
+        new_node = node.clone_item()
```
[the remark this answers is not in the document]
lnielsen: I think it comes from the [luqum docs example](https://luqum.readthedocs.io/en/latest/quick_start.html#manipulating) and that it's the intended way to do it: https://luqum.readthedocs.io/en/latest/api.html#luqum.tree.Item.clone_item

Issue #464, opened 2023-04-03
query parser: apply dependency constraints between constructor parameters


ppanero:
There are two more issues to take into account here:

**Dependency between `allow_list` and `mapping`**

When providing the `mapping` and the `allow_list`, we need to make sure that the first is at least a subset of the second. Meaning, all fields mapped should be part of the allow list, but this one can have some more if there's no need for mapping (e.g. identity function - email:email).  For example:

```python
mapping = {
    a: a.one,
    b: b.one,
}

allow_list = [a.one, b.one, another.one]
```

The following would be a misconfiguration:

```python
mapping = {
    a: a.one,
    b: b.one,
}

allow_list = [a.one, another.one]
```

One option could be to actually calculate the allow list from the mapping and enable injection of more fields, so we do not have to add all twice. Pseudocode:

```python
def __init__(mapping, allow_list):
    self.allow_list = set(allow_list.extend(mapping.values()))
    ...
```

**fields must be specified for `multi_field` queries**

The [QueryParser factory](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/records/queryparser/query.py#L71) must have a check for when the transformer returns an exception and defaults to a `multi_field`. In that case, the `fields` attribute must be defined (i.e. needs an assert).

It is unclear when to do this check if in the constructor, or in the except. Why? Because otherwise the multi-field will query all fields and we would be bypassing the allow list, and we could end up in an elasticsearch injection of some sorts and leaking data (e.g. user emails).

ppanero:
requires https://github.com/inveniosoftware/invenio-records-resources/issues/447 to be closed first

PR #465, opened 2023-04-05
queryparser: calculate mapping, allow list and fields

ppanero:
- closes https://github.com/inveniosoftware/invenio-records-resources/issues/464

### DIFF
--- a/invenio_records_resources/services/records/queryparser/query.py
+++ b/invenio_records_resources/services/records/queryparser/query.py
@@ -8,11 +8,13 @@
 
 """Lucene query syntax parser."""
 
+from copy import deepcopy
 from functools import partial
 
 from invenio_search.engine import dsl
 from luqum.exceptions import ParseError
 from luqum.parser import parser as luqum_parser
+from werkzeug.utils import cached_property
 
 from invenio_records_resources.services.errors import QuerystringValidationError
 
@@ -61,19 +63,57 @@ class QueryParser:
             )
     """
 
-    def __init__(self, identity=None, extra_params=None, tree_transformer_factory=None):
+    def __init__(self, identity=None, extra_params=None, tree_transformer_cls=None):
         """Initialise the parser."""
         self.identity = identity
-        self.extra_params = extra_params or {}
-        self.tree_transformer_factory = tree_transformer_factory
+        self.tree_transformer_cls = tree_transformer_cls
+        # the query parser is instantiated once per query and the extra params is a dict
+        # coming from a class attribute (passed by reference). then the popped attributes
+        # would disappear after one query. we need to pop to avoid passing them to the
+        # actual search query.
+        self.extra_params = deepcopy(extra_params) or {}
+        # the pop or {} is needed due to extra_params being passed from the factory
+        # it is possible that e.g. allow_list=None and then it will fail to set()
+        self.mapping = self.extra_params.pop("mapping", None) or {}
+        self._allow_list = self.extra_params.pop("allow_list", None)
+        # fields is not removed from extra params since if given it must be
+        # used in both querystring and multi match
+        self._fields = self.extra_params.get("fields") or []
+
+    @property
+    def allow_list(self):
+        """Calculate the allow list."""
+        if self._allow_list:  # only add the mapping if there is an allow list
+            # mapping should be at least a subset of the allow list
+            # implicit creation of the allow list has been chosen since there are cases
+            # were the mapping list has 10s or 100s of terms and we want to avoid
+            # having to duplicate those on creation in the allow list
+            return set(self._allow_list).union(self.mapping.values())
+        return set()
+
+    @cached_property
+    def fields(self):
+        """Calculate the list of fields to query.
+
+        It adds the allowed list of fields and remove duplications.
+        For example, boosted fields.
+        """
+        repeated = set()
+        for field in self._fields:  # check duplicated fields
+            field_name = field.split("^")[0]  # remove potential boosting
+            if field_name in self.allow_list:
+                repeated.add(field_name)
+
+        # fields = original fields + (allow list - repeated without boosting)
+        return list(set(self._fields).union(self.allow_list.difference(repeated)))
 
     @classmethod
-    def factory(cls, tree_transformer_factory=None, **extra_params):
+    def factory(cls, tree_transformer_cls=None, **extra_params):
         """Create a new instance of the query parser."""
         return partial(
             cls,
             extra_params=extra_params,
-            tree_transformer_factory=tree_transformer_factory,
+            tree_transformer_cls=tree_transformer_cls,
         )
 
     def parse(self, query_str):
@@ -82,13 +122,23 @@ class QueryParser:
             # We parse the Lucene query syntax in Python, so we know upfront
             # if the syntax is correct before executing it in the search engine
             tree = luqum_parser.parse(query_str)
-
             # Perform transformation on the abstract syntax tree (AST)
-            if self.tree_transformer_factory is not None:
-                transformer = self.tree_transformer_factory()
+            if self.tree_transformer_cls is not None:
+                transformer = self.tree_transformer_cls(
+                    mapping=self.mapping, allow_list=self.allow_list
+                )
                 new_tree = transformer.visit(tree, context={"identity": self.identity})
                 query_str = str(new_tree)
             return dsl.Q("query_string", query=query_str, **self.extra_params)
         except (ParseError, QuerystringValidationError):
             # Fallback to a multi-match query.
+            if self.allow_list:
+                # if there is an allow list it must overwrite a potential value
+                # given by the query to include it in the fields
+                kwargs = {**self.extra_params, "fields": self.fields}
+                return dsl.Q("multi_match", query=query_str, **kwargs)
+
+            # if there is no allow list we pass the parameters as default, without
+            # modifying the fields, or nothing if it was not passed. this is to
+            # avoid passing `fields=None`
             return dsl.Q("multi_match", query=query_str, **self.extra_params)
--- a/invenio_records_resources/services/records/queryparser/transformer.py
+++ b/invenio_records_resources/services/records/queryparser/transformer.py
@@ -30,11 +30,6 @@ class SearchFieldTransformer(TreeTransformer):
         self._allow_list = allow_list
         super().__init__(self, *args, **kwargs)
 
-    @classmethod
-    def factory(cls, mapping=None, allow_list=None):
-        """Create a new field transformer."""
-        return partial(cls, mapping or {}, allow_list or [])
-
     def visit_search_field(self, node, context):
         """Visit a search field."""
         # Use the node name if not mapped for transformation.
--- a/tests/services/test_service_queryparser.py
+++ b/tests/services/test_service_queryparser.py
@@ -14,7 +14,6 @@ from invenio_access.permissions import system_identity
 from invenio_records_resources.services.records.queryparser import (
     QueryParser,
     SearchFieldTransformer,
-    SuggestQueryParser,
 )
 
 
@@ -26,7 +25,7 @@ def parser():
 
 def test_extra_params():
     """Test for the query parser."""
-    p = QueryParser(system_identity).factory(fields=["title"])
+    p = QueryParser.factory(fields=["title"])
     assert p(system_identity).parse("a").to_dict() == {
         "query_string": {"query": "a", "fields": ["title"]}
     }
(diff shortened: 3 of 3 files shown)


tests/services/test_service_queryparser.py
```diff
@@ -104,16 +103,14 @@ def test_multimatch(parser, query):
         # Term is allowed and not mapped, results in a query string with the original query
         ("description:test", "description:test", "query_string", None),
         ("metadata.title:test", "metadata.title:test", "query_string", None),
-        # Terms might be mapped but not allowed, results in multi_match with the original query
-        ("title:test", "title:test", "multi_match", ["metadata.date"]),
```
ppanero, line 108: IMO, it does not make sense for the field to be mapped but not allowed. That means that the user can query it but cant at the same time. If something is defined in the mapping it will be allowed.

invenio_records_resources/services/records/queryparser/transformer.py
```diff
@@ -30,11 +30,6 @@ def __init__(self, mapping, allow_list, *args, **kwargs):
         self._allow_list = allow_list
         super().__init__(self, *args, **kwargs)
 
-    @classmethod
-    def factory(cls, mapping=None, allow_list=None):
```
ppanero, line 34: not needed anymore since now mapping and allow list are parameters of the query parser and passed down directly.

**IMPORTANT** this is a breaking change.

invenio_records_resources/services/records/queryparser/query.py
```diff
@@ -61,19 +61,55 @@ class SearchOptions:
...
+            # FIXME: it does not support fields with wildcards
+            field_name = field.split("^")[0]  # remove potential boosting
+            if field_name in self.allow_list:
+                to_remove.add(field_name)
+
+        # fields = original fields + (allow list - repeated without boosting)
+        return list(set(fields).union(self.allow_list.difference(to_remove)))
 
     @classmethod
-    def factory(cls, tree_transformer_factory=None, **extra_params):
```
ppanero, line 71: **IMPORTANT** this is a breaking change.

invenio_records_resources/services/records/queryparser/query.py
```diff
@@ -61,19 +61,55 @@ class SearchOptions:
...
         self.extra_params = extra_params or {}
-        self.tree_transformer_factory = tree_transformer_factory
+        # the pop or {} is needed due to extra_params being passed from the factory
+        # it is possible that e.g. allow_list=None and then it will fail to set()
+        self.mapping = self.extra_params.pop("mapping", None) or {}
+        self.allow_list = self._calculate_allow_list(
+            self.extra_params.pop("allow_list", None), self.mapping
+        )
+        # process fields list
+        self.fields = self._calculate_fields(
```
alejandromumo, line 76: small suggestion: does it make sense if `fields` is a cached property? In the current status, it's only used when the query defaults to `multi_match`.

invenio_records_resources/services/records/queryparser/query.py
```diff
@@ -82,13 +118,18 @@ def parse(self, query_str):
...
             # Perform transformation on the abstract syntax tree (AST)
-            if self.tree_transformer_factory is not None:
-                transformer = self.tree_transformer_factory()
+            if self.tree_transformer_cls is not None:
+                transformer = self.tree_transformer_cls(
+                    mapping=self.mapping, allow_list=self.allow_list
+                )
                 new_tree = transformer.visit(tree, context={"identity": self.identity})
                 query_str = str(new_tree)
             return dsl.Q("query_string", query=query_str, **self.extra_params)
```
alejandromumo, line 128: question: `extra_params` is passing the list of `fields` to be used in the query. Why do not pass `self.fields` directly given that we are computing them based on the allowed list? 

Same question applies to the multi match query below

invenio_records_resources/services/records/queryparser/query.py
ppanero: It can be made a `property`, that is true. However, I'm slightly against cached property, we've had some interesting side effects in the past (that's why I made it a field).

invenio_records_resources/services/records/queryparser/query.py
ppanero: The `fields` parameter is a bit special:
- If the parameter is given (as is in `extra_params`) it should be passed as is when using query string, to avoid biasing the boosting/filtering. That is why I'm not passing `self.fields`.
- In the case of `multi_field` query the calculated values should only be passed if there is an allowed list. Otherwise, we would overwrite a potential default or set it to `None`. That's why there is an `if` in the except block.

invenio_records_resources/services/records/queryparser/query.py
```diff
@@ -61,19 +61,55 @@ class SearchOptions:
...
+        """Calculate the list of fields to query."""
+        if not allow_list:
+            return fields
+
+        fields = fields or []
+        to_remove = set()
+        for field in fields:  # check duplicated fields
+            # FIXME: it does not support fields with wildcards
+            field_name = field.split("^")[0]  # remove potential boosting
+            if field_name in self.allow_list:
```
ntarocco, line 100: I think that here or we use the param (better) or we user `self.allow_list`:
```suggestion
            if field_name in allow_list:
```

invenio_records_resources/services/records/queryparser/query.py
```diff
@@ -61,19 +61,55 @@ class SearchOptions:
...
+        )
+        # process fields list
+        self.fields = self._calculate_fields(
+            # get instead of pop since fields could desired for
+            # both querystring and multi_match
+            self.extra_params.get("fields", None),
+            self.allow_list,
+        )
+
+    def _calculate_allow_list(self, allow_list, mapping):
```
ntarocco, line 83: does it make sense to have both method `static`, avoiding any temptation of using `self`?

invenio_records_resources/services/records/queryparser/query.py
```diff
@@ -61,19 +61,55 @@ class SearchOptions:
...
+        """Calculate the allow list."""
+        if allow_list:  # only add the mapping if there is an allow list
+            # mapping should be at least a subset of the allow list
+            return set(allow_list).union(mapping.values())
```
ntarocco, line 87: if I understand correctly, here we prefer calculating the allow list with the mappings, instead of failing if there is a mismatch, right?
I would honestly prefer a runtime failure (assert?), so that you know that they are misconfigured. I fear that, with this implementation, the behaviour will not be predictable, as the final list is computed merging the two.

It will indeed require being explicit and duplicate the list of fields.

So, instead, I would:
`assert mapping.values().issubset(set(allow_list)), raise ... a nice error msg ...`

invenio_records_resources/services/records/queryparser/query.py
```diff
@@ -61,19 +61,55 @@ class SearchOptions:
...
+
+    def _calculate_fields(self, fields, allow_list):
+        """Calculate the list of fields to query."""
+        if not allow_list:
+            return fields
+
+        fields = fields or []
+        to_remove = set()
+        for field in fields:  # check duplicated fields
+            # FIXME: it does not support fields with wildcards
```
ntarocco, line 98: This FIXME will be never fixed :) what happens if `*` is given? should we fail instead?

invenio_records_resources/services/records/queryparser/query.py
```diff
@@ -61,19 +61,55 @@ class SearchOptions:
...
+            field_name = field.split("^")[0]  # remove potential boosting
+            if field_name in self.allow_list:
+                to_remove.add(field_name)
+
+        # fields = original fields + (allow list - repeated without boosting)
+        return list(set(fields).union(self.allow_list.difference(to_remove)))
```
ntarocco, line 104: This is not clear to me: shouldn't it be only the list of allowed fields?

invenio_records_resources/services/records/queryparser/query.py
```diff
@@ -82,13 +118,18 @@ def parse(self, query_str):
...
+            if self.tree_transformer_cls is not None:
+                transformer = self.tree_transformer_cls(
+                    mapping=self.mapping, allow_list=self.allow_list
+                )
                 new_tree = transformer.visit(tree, context={"identity": self.identity})
                 query_str = str(new_tree)
             return dsl.Q("query_string", query=query_str, **self.extra_params)
         except (ParseError, QuerystringValidationError):
             # Fallback to a multi-match query.
+            if self.fields:
```
ntarocco, line 131: Why this `if self.fields`?

invenio_records_resources/services/records/queryparser/query.py
ppanero: It won't break but it will not work as expected. For example, it could be that the fields have `some.*` and then we have `some.else` in the allow list. Then for multi_match we would pass `some.*` and `some.else`, it could affect a bit the scoring by counting twice some field. I could dive on the outcome, but it won't break.

invenio_records_resources/services/records/queryparser/query.py
ppanero: On the one hand I agree, on the other hand it depends if we want to favor explicit or implicit configuration. We have three parameters affecting the allowed list: `fields`, `mapping` and `allowed list`. The reasons why I went implicit are:
- Some times the mapping is very very long, then we would need to duplicate a very long list, which is error prone to missing some. For example [ZenodoRDM](https://github.com/zenodo/zenodo-rdm/blob/master/invenio.cfg#L145).
- In my opinion it does not make sense to define a mapping, if you do not want to allow searching over it. Therefore, I see the allow list as an extra to "remove" the capabilities over fields that are not mentioned. For example: we want to map `user: user.username` and then remove the query over `email` (so we do not add it). It is true that this implementing a blacklist by omision.

invenio_records_resources/services/records/queryparser/query.py
ppanero: Related to the previous comments, I assumed that the user understand what its doing. If they boost a field, they consider that it can be queried. Mainly because on the `querystring` it will be passed no matter what (i.e. we can only block it in multi_field queries).

invenio_records_resources/services/records/queryparser/query.py
ppanero: I've refactored a bit the code to try to make it more understandable and included a comment in the code:

>  in the case that there is no allow list nor fields in the extra_params we do not enter this if to avoid passing `fields=None`, however, if it exists it must overwrite a potential value given by the query to include the allow_list

It is related to this comment/answer too, https://github.com/inveniosoftware/invenio-records-resources/pull/465#discussion_r1162537906

invenio_records_resources/services/records/queryparser/query.py
ppanero: Discussed IRL, implicit is okay due to long list of mappings.

invenio_records_resources/services/records/queryparser/query.py
```diff
@@ -61,19 +62,50 @@ class SearchOptions:
...
+        self.tree_transformer_cls = tree_transformer_cls
         self.extra_params = extra_params or {}
-        self.tree_transformer_factory = tree_transformer_factory
+        # the pop or {} is needed due to extra_params being passed from the factory
+        # it is possible that e.g. allow_list=None and then it will fail to set()
+        self.mapping = self.extra_params.pop("mapping", None) or {}
+        self._allow_list = self.extra_params.pop("allow_list", None)
+        # fields is not removed from extra params since if given it must be
+        # used in both querystring and multi match
+        self._fields = self.extra_params.get("fields", None)
```
zzacharo, line 76: nit:
```suggestion
        self._fields = self.extra_params.get("fields") or []
```

invenio_records_resources/services/records/queryparser/query.py
```diff
@@ -61,19 +62,50 @@ class SearchOptions:
...
+            # implicit creation of the allow list has been chosen since there are cases
+            # were the mapping list has 10s or 100s of terms and we want to avoid
+            # having to duplicate those on creation in the allow list
+            return set(self._allow_list).union(self.mapping.values())
+        return set()
+
+    @cached_property
+    def fields(self):
+        """Calculate the list of fields to query."""
+        fields = self._fields or []
```
zzacharo, line 92: with the suggestion above you can use directly `self._fields`

invenio_records_resources/services/records/queryparser/query.py
```diff
@@ -82,13 +114,23 @@ def parse(self, query_str):
...
+                )
                 new_tree = transformer.visit(tree, context={"identity": self.identity})
                 query_str = str(new_tree)
             return dsl.Q("query_string", query=query_str, **self.extra_params)
         except (ParseError, QuerystringValidationError):
             # Fallback to a multi-match query.
+            if self.allow_list:
+                # if there is an allow list it must overwrite a potential value
+                # given by the query to include it in the fields
+                kwargs = {**self.extra_params, "fields": self.fields}
```
zzacharo, line 130: that's different from `self._fields`?

ppanero: Yes, `self.fields` is a property that is calculated using the original `_fields`, adding the allowed list, and removing duplicates. Then depending on the case (i.e. the if in line 127) we might need to use the calculated fields or simply the raw original ones. That's why we do not modify them (they are still in extra_params).

The reason to use `_fields` and `fields` is to avoid doing the calculation in the constructor, since it is not always needed. See first review comments form manuel.

On top of that we extract from extra_params to a private attribute because we do not want to modify the values stored in extra params

zzacharo: It's dummy but could we rename it to `calculated_fields`? or have a detailed docstring on what are the differences between the `self.fields` and `self._fields`?

ppanero: I've added an explanation on the docstring and there is an explanation of the actual calculations in line 102

kpsherva (approved):

ntarocco (approved):

PR #466, opened 2023-04-18
resolvers: add ResultItemResolver

rekt-hard:
* closes https://github.com/inveniosoftware/invenio-notifications/issues/5

adapt parsing reference dict, as it may get an entity which does not inherit from dict

### DIFF
--- a/invenio_records_resources/references/entity_resolvers/base.py
+++ b/invenio_records_resources/references/entity_resolvers/base.py
@@ -19,19 +19,13 @@ def _parse_ref_dict(reference_dict, strict=True):
     The ``strict`` parameter controls if the number of keys in the
     reference dict is checked strictly or not.
     """
-    try:
-        keys = list(reference_dict.keys())
+    keys = list(reference_dict.keys())
 
-        if strict and len(keys) != 1:
-            raise ValueError(
-                "Reference dicts may only have one property! "
-                f"Offending dict: {reference_dict}"
-            )
-
-    except AttributeError:
-        # if reference dict is not really a dict
-        # happens if entity does not match the type and gets passed to _parse_ref_dict)
-        keys = None
+    if strict and len(keys) != 1:
+        raise ValueError(
+            "Reference dicts may only have one property! "
+            f"Offending dict: {reference_dict}"
+        )
 
     if not keys:
         return None
@@ -72,13 +66,11 @@ class EntityProxy(ABC):
 
     def _parse_ref_dict_type(self):
         """Parse the TYPE from the reference dict."""
-        r = _parse_ref_dict(self._ref_dict)
-        return r[0] if r else None
+        return _parse_ref_dict(self._ref_dict)[0]
 
     def _parse_ref_dict_id(self):
         """Parse the ID from the reference dict."""
-        r = _parse_ref_dict(self._ref_dict)
-        return r[1] if r else None
+        return _parse_ref_dict(self._ref_dict)[1]
 
     @property
     def reference_dict(self):
@@ -147,13 +139,11 @@ class EntityResolver(ABC):
 
     def _parse_ref_dict_type(self, ref_dict):
         """Parse the TYPE from the reference dict."""
-        r = self._parse_ref_dict(ref_dict)
-        return r[0] if r else None
+        return self._parse_ref_dict(ref_dict)[0]
 
     def _parse_ref_dict_id(self, ref_dict):
         """Parse the ID from the reference dict."""
-        r = self._parse_ref_dict(ref_dict)
-        return r[1] if r else None
+        return self._parse_ref_dict(ref_dict)[1]
 
     def get_entity_proxy(self, ref_dict, check=True):
         """Check compatibility and get a proxy for the referenced entity.
--- a/invenio_records_resources/references/entity_resolvers/base.py
+++ b/invenio_records_resources/references/entity_resolvers/base.py
@@ -19,13 +19,19 @@ def _parse_ref_dict(reference_dict, strict=True):
     The ``strict`` parameter controls if the number of keys in the
     reference dict is checked strictly or not.
     """
-    keys = list(reference_dict.keys())
+    try:
+        keys = list(reference_dict.keys())
 
-    if strict and len(keys) != 1:
-        raise ValueError(
-            "Reference dicts may only have one property! "
-            f"Offending dict: {reference_dict}"
-        )
+        if strict and len(keys) != 1:
+            raise ValueError(
+                "Reference dicts may only have one property! "
+                f"Offending dict: {reference_dict}"
+            )
+
+    except AttributeError:
+        # if reference dict is not really a dict
+        # happens if entity does not match the type and gets passed to _parse_ref_dict)
+        keys = None
 
     if not keys:
         return None
@@ -66,11 +72,13 @@ class EntityProxy(ABC):
 
     def _parse_ref_dict_type(self):
         """Parse the TYPE from the reference dict."""
-        return _parse_ref_dict(self._ref_dict)[0]
+        r = _parse_ref_dict(self._ref_dict)
+        return r[0] if r else None
 
     def _parse_ref_dict_id(self):
         """Parse the ID from the reference dict."""
-        return _parse_ref_dict(self._ref_dict)[1]
+        r = _parse_ref_dict(self._ref_dict)
+        return r[1] if r else None
 
     @property
     def reference_dict(self):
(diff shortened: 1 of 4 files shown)


invenio_records_resources/references/entity_resolvers/results.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def __init__(self, resolver, ref_dict, service):
+        """Constructor."""
+        super().__init__(resolver, ref_dict)
+        self.service = service
+
+    def _resolve(self):
+        """Resolve the result item from the proxy's reference dict."""
+        id_ = self._parse_ref_dict_id()
+        return self.service.read(system_identity, id_).to_dict()
```
slint, line 27: minor: not bocking, but could we make this identity configurable (i.e. passed from the `ResultItemResolver` to the `proxy_cls`)? It's a bit "dangerous" to have `system_identity` used like this since one might accidentally use it to get unauthorized information.

invenio_records_resources/references/entity_resolvers/results.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def get_needs(self, ctx=None):
+        """Return needs."""
+        # TODO: service.config.permission_policy_cls.can_read or []
+        #       Since resolving is done by the system, might not be applicable
+        return []
```
slint, line 33: comment: indeed in this case it's not needed since the `service.read` already applies the permission check (see also my comment on the configurable identity though)

invenio_records_resources/references/entity_resolvers/base.py
```diff
@@ -19,13 +19,19 @@ def _parse_ref_dict(reference_dict, strict=True):
...
+        if strict and len(keys) != 1:
+            raise ValueError(
+                "Reference dicts may only have one property! "
+                f"Offending dict: {reference_dict}"
+            )
+
+    except AttributeError:
+        # if reference dict is not really a dict
+        # happens if entity does not match the type and gets passed to _parse_ref_dict)
+        keys = None
```
slint, line 34: minor: I'm not sure about the exception handling here. Wouldn't it make sense to instead check if the `reference_dict` is actually a `dict` when initializing the proxy?

invenio_records_resources/references/entity_resolvers/base.py
```diff
@@ -66,11 +72,13 @@ def _parse_ref_dict(self):
 
     def _parse_ref_dict_type(self):
         """Parse the TYPE from the reference dict."""
-        return _parse_ref_dict(self._ref_dict)[0]
+        r = _parse_ref_dict(self._ref_dict)
+        return r[0] if r else None
```
slint, line 76: minor: it might be easier to just fail early in these cases, since I'm not sure if consumers of these methods take into account that they might return `None`. If `self._ref_dict` is not a `dict` maybe it's better to catch that in the constructor.

invenio_records_resources/references/entity_resolvers/results.py
```diff
@@ -0,0 +1,69 @@
...
+# details.
+
+"""Resolver for result items."""
+
+from invenio_access.permissions import system_identity
+
+from .base import EntityProxy, EntityResolver
+
+
+class ResultItemProxy(EntityProxy):
```
slint, line 16:
```suggestion
class ServiceResultProxy(EntityProxy):
```

minor: Technically, we're resolving to `ResultItem`s, but if we follow the naming convention of other `EntityProxy`s, they refer to the underlying entity/API.

invenio_records_resources/references/entity_resolvers/results.py
```diff
@@ -0,0 +1,69 @@
...
+        # TODO: service.config.permission_policy_cls.can_read or []
+        #       Since resolving is done by the system, might not be applicable
+        return []
+
+    def pick_resolved_fields(self, identity, resolved_dict):
+        """Select which fields to return when resolving the reference."""
+        return {"id": resolved_dict["id"]}
+
+
+class ResultItemResolver(EntityResolver):
```
slint, line 40:
```suggestion
class ServiceResultResolver(EntityResolver):
```

invenio_records_resources/references/entity_resolvers/results.py
```diff
@@ -0,0 +1,69 @@
...
+        :param type_key: The value to use for the TYPE part of the ref_dicts.
+        """
+        self.item_cls = item_cls
+        self.type_key = type_key
+        self.proxy_cls = proxy_cls
+        super().__init__(service_id)
+
+    def matches_entity(self, entity):
+        """Check if the entity is a result item."""
+        return isinstance(entity, self.item_cls)
```
slint, line 57:
```suggestion
        return isinstance(entity, (self.item_cls, self.get_service().record_cls))
```

minor: so that we can call this inside the service calls (that don't have access to a `ResultItem` yet, but probably the underlying API class)

slint (commented):
LGTM, one shelvable comment on the identity used for the proxy, some naming, and minor comment on skipping a check.

invenio_records_resources/references/entity_resolvers/base.py
rekt-hard: It's about these lines here:

https://github.com/inveniosoftware/invenio-records-resources/blob/ec55cf197ddfc5a0231818345dbc8822573b2201/invenio_records_resources/references/registry.py#L83-L91

This works for classes which inherit from `Record`, since this is also a dict. Then only a `ValueError` is raised. If another class is passed here (e.g. `ResultItem`), then an `AttributeError` will be raised.

So it could be ok to just add the `AttributeError` to the except block, so this is caught as well.

invenio_records_resources/references/entity_resolvers/base.py
rekt-hard: This was also just added because of the dict inheritance of record classes, which others might not have. So it can be removed, when the other issue is resolved

invenio_records_resources/references/entity_resolvers/results.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def __init__(self, item_cls, service_id, type_key, proxy_cls=ResultItemProxy):
+        """Constructor.
+
+        :param item_cls: The result item class to use.
+        :param service_id: The record service id.
+        :param type_key: The value to use for the TYPE part of the ref_dicts.
+        """
+        self.item_cls = item_cls
+        self.type_key = type_key
```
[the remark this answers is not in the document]
rekt-hard: Yes, good idea!
In most cases, the service id will be in plural (e.g. `communities`, `requests`). Since the same value is used for the ref dict and when resolving, it should be fine.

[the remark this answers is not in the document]
slint: Hm, now that I see it actually, the plural breaks a bit the intuitiveness of `{"request": "<req_id>"}` vs. `{"requests": "<req_id>"}`. Let's keep the `type_key`/`type_id` pattern, and get only the `item_cls` from the service config.

invenio_records_resources/references/entity_resolvers/base.py
slint: Ok I see now. We should only aim at calling `.reference_entity(...)` on API-class/Record instances and not `ResultItem` instances. I'm a bit worried about `AttributeError` masking e.g. an actual bug in the underlying code, in which case we don't have a way to differentiate.

I also suspect that the `ValueError` here is aimed at catching the exception raised at https://github.com/inveniosoftware/invenio-records-resources/blob/58503e2a6fdc7bd110f7d3bf3a51e7ca05fd63f0/invenio_records_resources/references/entity_resolvers/base.py#L167 which is raised/controlled by us.

rekt-hard: Do you mean `reference_entity(...)` or `matches_reference_dict(...)` should only be called on API-class/Record instances?
If second, a type check of the entity could work
```py
elif isinstance(entity, dict) and resolver.matches_reference_dict(entity):
  return entity```

So probably what you wrote in your first comment? :)

Comment (rekt-hard) [reply invenio_records_resources/references/entity_resolvers/base.py]: Updated it to add this check

File: invenio_records_resources/references/entity_resolvers/results.py [hunk 9]
```diff
@@ -0,0 +1,76 @@
...
+
+        :param item_cls: The result item class to use.
+        :param service_id: The record service id.
+        :param type_key: The value to use for the TYPE part of the ref_dicts.
+        """
+        super().__init__(service_id)
+        self.type_key = type_key
+        self.proxy_cls = proxy_cls
+        self.item_cls = item_cls or self.get_service().config.result_item_cls
+        self.record_cls = record_cls or self.get_service().record_cls
```
slint, line 60:
```suggestion
        self.item_cls = item_cls or self.get_service().config.result_item_cls
        self.record_cls = record_cls or self.get_service().record_cls
```

minor: we should compute these "lazily" either in the functions they're used in or via properties, since depending on when we initialize the `ServiceResultResolver` not all services/registries will be initialized.

slint (approved):
LGTM, one last comment on having `item_cls` and `record_cls` being lazy, and we can merge. I'll split https://github.com/inveniosoftware/invenio-records-resources/pull/466/files#r1176103367 into a separate issue.

invenio_records_resources/references/entity_resolvers/results.py
slint: Discussed IRL, and this is fine for now.

slint (approved):
LGTM, I think we can merge :)

kpsherva (approved):

PR #470, opened 2023-05-12
expanded fields: add a non-resolvable system record

anikachurilova:
* introduce a base ghost schema
* closes https://github.com/zenodo/zenodo-rdm/issues/284


invenio_records_resources/services/records/schema.py
```diff
@@ -42,6 +42,14 @@ def clean(self, data, **kwargs):
         return data
 
 
+class BaseGhostSchema(Schema):
+    """Base ghost schema."""
+
+    SYSTEM_USER_ID = "-1"
```
anikachurilova, line 48: Not sure about putting it here. The best would be to place it in the [SystemUserSchema](https://github.com/inveniosoftware/invenio-users-resources/pull/74/files#diff-503368b315245f66e0a168f6ce6a972b3e32f4258b1b6d3980b36bd3730b8941R98), but then this constant can't be used in `invenio-records-resources` as it doesn't import `invenio-users-resources` at all (it happens the other way around and created the loop import)

invenio_records_resources/services/records/results.py
```diff
@@ -292,10 +294,12 @@ def add_service_value(self, service, value):
 
     def add_dereferenced_record(self, service, value, resolved_rec):
         """Save the dereferenced record."""
+        # mark the record as a "ghost" or "system" record i.e not resolvable
         if resolved_rec is None:
-            resolved_rec = self.ghost_record({"id": value})
-            # mark the record as a "ghost" record i.e not resolvable
-            resolved_rec["is_ghost"] = True
+            if value == system_user_id:
+                resolved_rec = self.system_record()
```
alejandromumo, line 300: where does this `system_record` come from?

invenio_records_resources/services/records/results.py
```diff
@@ -292,10 +294,12 @@ def add_service_value(self, service, value):
...
+            else:
+                resolved_rec = self.ghost_record({"id": value})
```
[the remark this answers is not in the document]
anikachurilova: I see it removed here on line 298, do you see it somewhere else?

invenio_records_resources/services/records/results.py
anikachurilova: Yes, but then all the classes that inherit from will have to implement the `system_record()`, which doesn't make sense for them (for example `ParentCommunitiesExpandableField`)

alejandromumo: Then, can `self.system_record()` be undefined at some point?
You can also declare it without being abstract and it doesn't do anything (no-op). It's then overriden by whoever needs to be override it.

invenio_records_resources/services/records/results.py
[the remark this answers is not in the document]
alejandromumo: Not really, my question was whether its removal was a leftover or intended 😛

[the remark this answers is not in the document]
anikachurilova: Ah, then yes, the `is_ghost` will always be `true` as I made it a constant in a parent `BaseGhostSchema` class (see below)

invenio_records_resources/services/records/results.py
anikachurilova: Good point! Feels like better to make it fail here with the  `NotImplementedError` (same as the `ghost_record ` above, but make it not abstract), so that we know where and why it failed, WDYT?
ping @kpsherva

kpsherva: 👍  on following the ghost

PR #471, opened 2023-05-23
pids: add resolver for pid

TLGINO:
* closes inveniosoftware/invenio-rdm-records#1282

### DIFF
--- a/invenio_records_resources/services/records/service.py
+++ b/invenio_records_resources/services/records/service.py
@@ -359,10 +359,25 @@ class RecordService(Service, RecordIndexerMixin):
             expand=expand,
         )
 
-    def read(self, identity, id_, expand=False):
+    def read(self, identity, id_, expand=False, front_end_call=False):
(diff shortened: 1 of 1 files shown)


invenio_records_resources/services/records/service.py
```diff
@@ -359,10 +359,25 @@ def _create(
...
+    def read(self, identity, id_, expand=False, front_end_call=False):
         """Retrieve a record."""
         # Resolve and require permission
-        record = self.record_cls.pid.resolve(id_)
+        from sqlalchemy.orm.exc import NoResultFound
+
+        try:
+            record = self.record_cls.pid.resolve(id_)
+        except NoResultFound as resolved_pid:
+            hits = self.search(identity, q=f"parent.id:{id_}")
```
ntarocco, line 370: We should not use a search to retrieve records in the `read` method: IMO, we should prefer accessing the DB as source of truth, rather then the search engine.

PR #472, opened 2023-05-31
Expandable fields: Skip expansion if the value or service is None

max-moser:
Some entities are abstract and not backed by a service, like system roles.

It might be however that we encounter such entities during expansion, among others.
One such case is the expansion process for access grants, which allows the referenced entities to be users (:heavy_check_mark: ), roles (:heavy_check_mark:) or system roles (:x:).

This PR chooses to ignore/skip entities that don't have a service with which to resolve them, because that's simpler than adding code in other places that checks for entity types explicitly before trying to dump result lists with expansion on.

### DIFF
--- a/invenio_records_resources/services/records/results.py
+++ b/invenio_records_resources/services/records/results.py
@@ -344,7 +344,7 @@ class FieldsResolver:
 
         :params expandable_fields: list of ExpandableField obj.
         """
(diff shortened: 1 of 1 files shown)


max-moser:
Added a dummy service in RDM-Records for the entities in question, so this isn't necessary anymore.

PR #478, opened 2023-07-31
relations: reindex by chunk

ntarocco:
* the reindex of records on relation update was not limiting the number of records given as input. For large number of records, the search engine fails because of too many search clauses added to the search query..


slint (approved):
LGTM! One concern I have is what happens if one of the reindex chunks fails in the middle, but we weren't in a better state before either (i.e. the whole batch would just fail).

PR #481, opened 2023-08-08
params: refactored sort fields computation.

alejandromumo:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1398

WHY THIS PR:

- I need to implement a type of `SortParam`, where the `fields` are computed in the same way as the parent except that, in the new interpreter, the fields are always prepended by a new field.
- To avoid code duplication, I refactored the `apply` method to just apply the fields to the search's sort instead of computing the fields + applying them.
- Added a service call to reindex records by the latest version first. 
  - The idea is to speed up the reindexing of the latest versions while the other versions can be done, in bulk, separately. 
- Added a new UOW operation for bulk indexing records, given an iterable of ids (e.g. a generator)

### DIFF
--- a/invenio_records_resources/services/records/service.py
+++ b/invenio_records_resources/services/records/service.py
@@ -23,7 +23,13 @@ from invenio_records_resources.services.errors import PermissionDeniedError
 
 from ..base import LinksTemplate, Service
 from ..errors import RevisionIdMismatchError
-from ..uow import RecordCommitOp, RecordDeleteOp, unit_of_work
+from ..uow import (
+    RecordBulkIndexOp,
+    RecordCommitOp,
+    RecordDeleteOp,
+    RecordIndexOp,
+    unit_of_work,
+)
 from .schema import ServiceSchemaWrapper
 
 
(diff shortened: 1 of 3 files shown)


invenio_records_resources/services/records/service.py
```diff
@@ -552,3 +558,37 @@ def on_relation_update(
...
+    @unit_of_work()
+    def reindex_latest_first(
+        self, identity, search_preference=None, extra_filter=None, uow=None
+    ):
+        """Reindexes records matching the query filter, prioritizing latest versions.
+
+        Records will be retrieved from search engine and current versions re-indexed first.
+        The latest version of a record is reindexed twice, but we trade that off for the guarantee that it also gets re-indexed first.
+        """
+        self.require_permission(identity, "manage")
```
ntarocco, line 571: Minor: the permission of `reindex` is `search`, and not `manage`. We might want to align the 2? `manage` actually looks better, because we should re-index when something changed. The only exception is the `relation update`: when a related record is updated, it can trigger a re-index (probably with system process).

invenio_records_resources/services/records/service.py
```diff
@@ -552,3 +558,37 @@ def on_relation_update(
 
             self.reindex(identity, search_query=search_query)
         return True
+
+    @unit_of_work()
+    def reindex_latest_first(
+        self, identity, search_preference=None, extra_filter=None, uow=None
+    ):
+        """Reindexes records matching the query filter, prioritizing latest versions.
```
ntarocco, line 566: I would add as a comment that this should happen in a celery task, and not in sync request, given that `scan` might be slow.

invenio_records_resources/services/records/service.py
```diff
@@ -552,3 +558,37 @@ def on_relation_update(
...
+            identity,
+            self.record_cls,
+            self.config.search,
+            permission_action="read",
+            extra_filter=extra_filter,
+            preference=search_preference,
+        )
+
+        # Search only latest versions and register them for indexing
+        latest_versions = search.filter("term", **{"versions.is_latest": True}).scan()
```
ntarocco, line 584: Do you want to be extra safe and handle any `NoResultFound` exception? Maybe simply `continue` in such cases?

invenio_records_resources/services/records/service.py
```diff
@@ -552,3 +558,37 @@ def on_relation_update(
...
+        for res in latest_versions:
+            record = self.record_cls.pid.resolve(res.id)
+            uow.register(RecordIndexOp(record, indexer=self.indexer))
+
+        # Search all versions and register them for bulk indexing
+        all_versions = search.source(False).scan()
```
ntarocco, line 590: I would have added a comment, I didn't recall what `source` does :)
```suggestion
        # source(False): don't return any field, just the metadata
        all_versions = search.source(False).scan()
```

ntarocco (approved):

invenio_records_resources/services/records/service.py
alejandromumo: I had this question myself. However, since we are allowing the reindexing of records (any records) I thought it would be better to have it more restricted.  In reality, the `search` object will only allow the caller to reindex records that they can `read`, so it wouldn't be too bad to allow e.g. a user to reindex their own records.

ntarocco (approved):

PR #492, opened 2023-08-31
files: add flag to dump files count

zzacharo:
closes https://github.com/zenodo/rdm-project/issues/166

### DIFF
--- a/invenio_records_resources/records/systemfields/files/field.py
+++ b/invenio_records_resources/records/systemfields/files/field.py
@@ -120,8 +120,8 @@ class FilesField(SystemField):
 
         self.store(record, files)
 
-    def post_dump(self, record, data, **kwargs):
-        """Called after a record is dumped in a secondary storage system."""
+    def pre_dump(self, record, data, **kwargs):
+        """Called before a record is dumped in a secondary storage system."""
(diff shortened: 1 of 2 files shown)


invenio_records_resources/records/systemfields/files/field.py
```diff
@@ -225,7 +225,9 @@ def dump(self, record, files, include_entries=False):
             data["default_preview"] = files.default_preview
 
         if include_entries and files.enabled:
-            data["count"] = len(files)
+            if dump_count:
+                # needed because of https://github.com/zenodo/rdm-project/issues/166
+                data["count"] = len(files)
```
slint, line 230: The issue with exposed files data for restricted records is not only about the file count, but all attributes dumped here. In general, for restricted/closed-access files we should just not dump anything about the files content at all. Even the `default_preview` and `order` is somewhat "leaking" information about the files and contents.

zzacharo: After investigation, I am leaning towards making the `self._dump` property callable so it can take into account the record. The problem is that a search dumper is called before the `FilesField.post_dump` method where the dump is happening based on some flags. It would be much bigger change to move everything to a dumper potentially. If there is no strong opinion against I would go with that solution

PR #498, opened 2023-09-06
service: add sort to read_many

jrcastro2:
* closes https://github.com/zenodo/who-rdm/issues/82

### DIFF
--- a/invenio_records_resources/services/records/service.py
+++ b/invenio_records_resources/services/records/service.py
@@ -398,6 +398,7 @@ class RecordService(Service, RecordIndexerMixin):
         search_opts=None,
         extra_filter=None,
         preference=None,
+        sort=None,
         **kwargs,
     ):
         """Search for records matching the ids."""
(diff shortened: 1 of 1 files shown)


invenio_records_resources/services/records/service.py
```diff
@@ -398,6 +398,7 @@ def _read_many(
         search_opts=None,
         extra_filter=None,
         preference=None,
+        sort=None,
```
kpsherva, line 401: isn't sort passed in search_opts ? if you add it to the search configuration on the correct service (check RDMSearchOptions for reference) it should work without adding this kwarg

zzacharo: Agreed to keep the existing solution as the alternative would require some refactoring between the different methods of searching like search, read_many, read_all.... Vocabularies are using directly the `_read_many()` method.

zzacharo (approved):

Issue #56, opened 2020-07-29
Make maximum number of results configurable


fenekku:
(maybe this doesn't need to be configurable if ES caps at 10000)

This could be configured in invenio-records-rest, so maybe we want to be able to configure this still.

(honestly I am not a particularly big proponent of this)

PR #511, opened 2023-09-20
service: display tombstone for deleted record's files

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1479

The following endpoints display the tombstone page:
- `api/records/<pid>/files`
- `api/records/<pid>/files/<key>`
- `api/records/<pid>/media-files`
- `api/records/<pid>/files/<key>/content`
- `api/records/<pid>/files-archive
`

### DIFF
--- a/invenio_records_resources/resources/errors.py
+++ b/invenio_records_resources/resources/errors.py
@@ -9,11 +9,14 @@
 # details.
 
 """Common Errors handling for Resources."""
+import json
 from json import JSONDecodeError
 
 import marshmallow as ma
-from flask import jsonify, make_response, request, url_for
-from flask_resources import HTTPJSONException, create_error_handler
+from flask import g, jsonify, make_response, request, url_for
+from flask_resources import HTTPJSONException as _HTTPJSONException
+from flask_resources import create_error_handler
+from flask_resources.serializers.json import JSONEncoder
 from invenio_i18n import lazy_gettext as _
 from invenio_pidstore.errors import (
     PIDAlreadyExists,
@@ -36,11 +39,12 @@ from ..services.errors import (
     FileKeyNotFoundError,
     PermissionDeniedError,
     QuerystringValidationError,
+    RecordDeletedException,
     RevisionIdMismatchError,
 )
 
 
-class HTTPJSONValidationException(HTTPJSONException):
+class HTTPJSONValidationException(_HTTPJSONException):
     """HTTP exception serializing to JSON and reflecting Marshmallow errors."""
 
     description = "A validation error occurred."
@@ -50,7 +54,7 @@ class HTTPJSONValidationException(HTTPJSONException):
         super().__init__(code=400, errors=validation_error_to_list_errors(exception))
 
 
-class HTTPJSONSearchRequestError(HTTPJSONException):
+class HTTPJSONSearchRequestError(_HTTPJSONException):
     """HTTP exception responsible for mapping search engine errors."""
 
     causes_responses = {
@@ -68,6 +72,37 @@ class HTTPJSONSearchRequestError(HTTPJSONException):
                 super().__init__(code=code, description=msg)
                 return
         super().__init__(code=500, description=_("Internal server error"))
+
+
+class HTTPJSONException(_HTTPJSONException):
+    """HTTPJSONException that supports setting some extra body fields."""
+
+    def __init__(self, code=None, errors=None, **kwargs):
+        """Constructor."""
+        description = kwargs.pop("description", None)
+        response = kwargs.pop("response", None)
+        super().__init__(code, errors, description=description, response=response)
+        self._extra_fields = kwargs
+
+    def get_body(self, environ=None, scope=None):
+        """Get the response body."""
+        body = {
+            "status": self.code,
+            "message": self.get_description(environ),
+            **self._extra_fields,
+        }
+
+        errors = self.get_errors()
+        if errors:
+            body["errors"] = errors
+
+        # TODO: Revisit how to integrate error monitoring services.
+        # See issue https://github.com/inveniosoftware/flask-resources/issues/56
+        # Temporarily kept for expediency and backward-compatibility
+        if self.code and (self.code >= 500) and hasattr(g, "sentry_event_id"):
+            body["error_id"] = str(g.sentry_event_id)
+
+        return json.dumps(body, cls=JSONEncoder)
 
 
 def create_pid_redirected_error_handler():
@@ -185,5 +220,16 @@ class ErrorHandlersMixin:
         ),
         search.exceptions.RequestError: create_error_handler(
             lambda e: HTTPJSONSearchRequestError(e)
+        ),
+        RecordDeletedException: create_error_handler(
+            lambda e: (
+                HTTPJSONException(code=404, description=_("Record not found"))
+                if not e.record.tombstone.is_visible
+                else HTTPJSONException(
+                    code=410,
+                    description=_("Record deleted"),
+                    tombstone=e.record.tombstone.dump(),
+                )
+            )
         ),
     }
--- a/invenio_records_resources/resources/files/config.py
+++ b/invenio_records_resources/resources/files/config.py
@@ -11,6 +11,8 @@
 
 from flask_resources import ResourceConfig
 
+from invenio_records_resources.resources.files.args import RequestExtraArgsSchema
+
 
 class FileResourceConfig(ResourceConfig):
     """File resource config."""
@@ -27,3 +29,5 @@ class FileResourceConfig(ResourceConfig):
         "item-commit": "/files/<key>/commit",
         "list-archive": "/files-archive",
     }
+
+    request_extra_args = RequestExtraArgsSchema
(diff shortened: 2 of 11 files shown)


invenio_records_resources/resources/errors.py
```diff
@@ -70,6 +74,36 @@ def __init__(self, error):
         super().__init__(code=500, description=_("Internal server error"))
 
 
+class HTTPJSONException(_HTTPJSONException):
```
jrcastro2, line 77: this was moved here from invenio-rdm-records as it's required in this module as well

invenio_records_resources/services/errors.py
```diff
@@ -69,3 +69,12 @@ def __init__(self, recid, file_key):
         )
         self.recid = recid
         self.file_key = file_key
+
+
+class RecordDeletedException(Exception):
```
jrcastro2, line 74: this was moved here from invenio-rdm-records as it's required in this module as well

invenio_records_resources/services/files/service.py
```diff
@@ -68,10 +68,11 @@ def _get_record(self, id_, identity, action, file_key=None):
     #
     # High-level API
     #
-    def list_files(self, identity, id_):
+    def list_files(self, identity, id_, with_deleted=False):
```
zzacharo, line 71: are all the `with_deleted` params being checked against permissions? I guess we need a `<action>_deleted` permission to check if you can perform the action?

zzacharo: check how comment here https://github.com/inveniosoftware/invenio-rdm-records/pull/1489/files/492f2201d24152f813a5b8740a6fb64518753717#diff-d79cbaf668935422b2743d81ca9906d9b72ac8b3456c2280ba5a5edacb72796bR387

invenio_records_resources/resources/files/config.py
```diff
@@ -27,3 +28,5 @@ class FileResourceConfig(ResourceConfig):
         "item-commit": "/files/<key>/commit",
         "list-archive": "/files-archive",
     }
+
+    request_extra_args = {"include_deleted": ma.fields.Boolean()}
```
kpsherva, line 32: this is usually configured as schema, not dict, see https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/resources/args.py

invenio_records_resources/records/api.py
```diff
@@ -63,11 +63,11 @@ def get_by_key(cls, record_id, key):
                 return cls(obj.data, model=obj)
 
     @classmethod
-    def list_by_record(cls, record_id, with_deleted=False):
+    def list_by_record(cls, record_id, include_deleted=False):
```
kpsherva, line 66: unfortunately this might cause issues.
the with_deleted is coming from invenio_records and I think we are now hiding this functionality

invenio_records_resources/records/api.py
```diff
@@ -63,11 +63,11 @@ def get_by_key(cls, record_id, key):
...
         """List all record files by record ID."""
         query = cls.model_cls.query.filter(cls.model_cls.record_id == record_id)
 
-        if not with_deleted:
+        if not include_deleted:
```
kpsherva, line 70: query below applies for records which were deleted via .delete method of invenio-records api.
the records we delete are deleted by .delete_record of invenio-rdm-records, this query will cause issues, probably if you delete records in a new way, the files still show up on the list

invenio_records_resources/records/api.py
kpsherva: see also (got it while testing):
Screenshot 2023-09-25 at 17 02 04

jrcastro2: The error seems to be due to missing the changes from invenio-app-rdm, probably you are missing the changes of the [following PR ](https://github.com/inveniosoftware/invenio-app-rdm/pull/2434). If not, please describe how can I reproduce it and I will fix it.

invenio_records_resources/resources/errors.py
```diff
@@ -70,6 +74,36 @@ def __init__(self, error):
...
+            "status": self.code,
+            "message": self.get_description(environ),
+            **self._extra_fields,
+        }
+
+        errors = self.get_errors()
+        if errors:
+            body["errors"] = errors
+
+        # TODO: Revisit how to integrate error monitoring services. See issue #56
```
ntarocco, line 99: What is the issue #56?
It is not easy to understand well what is the issue here.

invenio_records_resources/services/files/service.py
```diff
@@ -65,13 +65,27 @@ def _get_record(self, id_, identity, action, file_key=None):
 
         return record
 
+    def _check_record_deleted(self, record, identity, include_deleted):
+        """Checks if record is deleted."""
```
ntarocco, line 69: Very minor: reading `check something` I would expect that the function returns True or False. What about this?
```suggestion
    def _ensure_record_not_deleted(self, record, identity, include_deleted):
        """Ensure that the record exists (not deleted) or raise."""
```

invenio_records_resources/resources/errors.py
jrcastro2: Right! Thank you for pointing this out. This code was moved here from [invenio-rdm-records](https://github.com/inveniosoftware/invenio-rdm-records/pull/1490/files#diff-4e526346ebe67a7c163c6811b36f9b13a88fbc8058ba43b0b1fba00e89fa297eL50). And apparently that was copied from flask-resources, I have updated the comment to point to the correct [issue](https://github.com/inveniosoftware/flask-resources/issues/56)

invenio_records_resources/resources/files/resource.py
```diff
@@ -84,13 +85,16 @@ def create_url_rules(self):
             ]
         return url_rules
 
+    @request_extra_args
```
kpsherva, line 88: any reason why you chose the extra_args instead of view args ? I did the opposite actually in rdm-records and I am considering changing

tests/services/files/test_file_service.py
```diff
@@ -450,3 +452,108 @@ def test_read_not_committed_external_file(
...
+            },
+        }
+    ]
+    # Initialize file saving
+    result = file_service.init_files(identity_simple, recid, file_to_initialise)
+    assert result.to_dict()["entries"][0]["key"] == file_to_initialise[0]["key"]
+    # for to_file in to_files:
+    content = BytesIO(b"test file content")
+    result = file_service.set_file_content(
+        identity_simple,
```
kpsherva, line 491: normally I would say you should not be able to set_file_content on deleted record unless for some reason the admin's job requires it. I would ask the zenodo support team for their use cases
I see two options
1. this endpoint should be protected by permissions, and only admin should be able to set content/commit - in this case we miss a test to check if normal user or owner of the record still can commit files
2. nobody, even the admin/moderator (except superuse) should be able to commit or set content

tests/services/files/test_file_service.py
```diff
@@ -450,3 +452,108 @@ def test_read_not_committed_external_file(
...
+            identity_simple, recid, "article.txt", include_deleted=True
+        )
+
+    result = file_service.get_file_content(
+        superuser_identity, recid, "article.txt", include_deleted=True
+    )
+    assert result.file_id == "article.txt"
+
+    # Delete file
+    result = file_service.delete_file(identity_simple, recid, "article.txt")
```
kpsherva, line 548: do we allow to an user to delete a file from deleted record? I don't think it should happen. Once the record is deleted, it should be "frozen", check if RFC has any info about this case: https://github.com/inveniosoftware/invenio-app-rdm/issues/2289

invenio_records_resources/services/files/service.py
```diff
@@ -65,13 +65,27 @@ def _get_record(self, id_, identity, action, file_key=None):
 
         return record
 
+    def _ensure_record_not_deleted(self, record, identity, include_deleted):
+        """Ensure that the record exists (not deleted) or raise."""
+        if record.is_draft:
```
kpsherva, line 70: I am not sure about this part here, we are in records-resouces, we shouldn't know about drafts' existence in this module

tests/services/files/test_file_service.py
jrcastro2: This feels a bit out of scope for this [issue](https://github.com/inveniosoftware/invenio-rdm-records/issues/1479), the goal is to hide the files of deleted records. The behaviour of who can update files of deleted records was not changed, at the moment it's only admins. 

If we want to change this we should add the tests in `invenio-rdm-records` as here we mocking the permissions.

tests/services/files/test_file_service.py
jrcastro2: We are using mocked [permissions](https://github.com/inveniosoftware/invenio-records-resources/blob/master/tests/mock_module/permissions.py) here, that's why a simple identity can delete files, the real permissions set for rdm should be tested in invenio-rdm-records.

invenio_records_resources/services/files/service.py
jrcastro2: I changed it to be `record.is_published`

PR #541, opened 2023-11-01
manager: update delete_all

jrcastro2:
* closes https://github.com/zenodo/rdm-project/issues/413
* NEEDS https://github.com/inveniosoftware/invenio-files-rest/pull/298

### DIFF
--- a/invenio_records_resources/records/api.py
+++ b/invenio_records_resources/records/api.py
@@ -100,6 +100,12 @@ class FileRecord(RecordBase, SystemFieldsMixin):
         """Get the file's record."""
         return self.record_cls(self._record.data, model=self._record)
 
+    @classmethod
+    def remove_all(cls, record_id):
+        """Hard delete record's file instances."""
+        record_files = cls.model_cls.query.filter(cls.model_cls.record_id == record_id)
+        record_files.delete()
+
     send_signals = False
     enable_jsonref = False
 
--- a/invenio_records_resources/records/systemfields/files/manager.py
+++ b/invenio_records_resources/records/systemfields/files/manager.py
@@ -249,6 +249,22 @@ class FilesManager(MutableMapping):
                 remove_rf=remove_rf,
             )
 
+    def teardown(self, full=True):
+        """Clean up all file manager related instances.
+
+        Deletes FileRecords, Bucket and ObjectVersions attached to file manager.
+        equivalent of:
+        softdelete_obj = false, remove_rf=True, remove_obj=true
+        option of deleting only FileRecords if full=False
+        """
+        self.file_cls.remove_all(self.record.id)
+
+        if full:
+            self.remove_bucket(force=True)
+        self.default_preview = None
+        self._entries = None
+        self._order = []
+
     def copy(self, src_files, copy_obj=True):
         """Copy from another file manager."""
         self.enabled = src_files.enabled
--- a/tests/records/test_systemfield_files.py
+++ b/tests/records/test_systemfield_files.py
@@ -9,7 +9,6 @@
 
 """Files field tests."""
 
-
 from io import BytesIO
 
 from invenio_files_rest.models import Bucket, FileInstance, ObjectVersion
@@ -133,7 +132,7 @@ def test_record_files_operations(base_app, db, location):
 
 
 def test_record_files_clear(base_app, db, location):
-    """Test clearing record files."""
+    """Test clearing record files (hard deletion)."""
     record = Record.create({})
 
     # Add a file with metadata + bytes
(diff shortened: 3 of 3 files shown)


invenio_records_resources/records/systemfields/files/manager.py
```diff
@@ -241,13 +241,27 @@ def delete(self, key, remove_obj=True, softdelete_obj=True, remove_rf=False):
...
-                softdelete_obj=softdelete_obj,
-                remove_rf=remove_rf,
-            )
+            rf = self[key]
+            if rf.object_version:
+                object_versions.append(rf.object_version)
+            # Remove or softdelete the entire row
+            rf.delete(force=remove_rf)
+
+        if remove_obj:
```
kpsherva, line 252: I think this should be handled in self.delete above, not here

invenio_records_resources/records/systemfields/files/manager.py
```diff
@@ -241,13 +241,27 @@ def delete(self, key, remove_obj=True, softdelete_obj=True, remove_rf=False):
...
         """Delete all file records."""
+        object_versions = []
         for key in list(self.keys()):
-            self.delete(
-                key,
-                remove_obj=remove_obj,
-                softdelete_obj=softdelete_obj,
-                remove_rf=remove_rf,
-            )
+            rf = self[key]
```
kpsherva, line 246:
```suggestion
            record_file = self[key]
```
for readability

invenio_records_resources/records/systemfields/files/manager.py
```diff
@@ -241,13 +241,27 @@ def delete(self, key, remove_obj=True, softdelete_obj=True, remove_rf=False):
...
+            if rf.object_version:
+                object_versions.append(rf.object_version)
+            # Remove or softdelete the entire row
+            rf.delete(force=remove_rf)
+
+        if remove_obj:
+            if softdelete_obj:
+                for object_version in object_versions:
+                    ObjectVersion.delete(object_version.bucket, object_version.key)
```
kpsherva, line 255: do I understand correctly that soft delete is still handled on one by one basis ?

invenio_records_resources/records/systemfields/files/manager.py
```diff
@@ -241,13 +241,27 @@ def delete(self, key, remove_obj=True, softdelete_obj=True, remove_rf=False):
...
+            else:
+                version_ids = [
+                    object_version.version_id for object_version in object_versions
+                ]
+                ObjectVersion.remove_many(version_ids, self.bucket)
```
kpsherva, line 260: ok I see now why you called it remove_many now, please ignore the comment in the other PR

lnielsen:
Wait a bit to merge, let's have a quick discussion on this one.

invenio_records_resources/records/api.py
```diff
@@ -100,6 +100,12 @@ def record(self):
         """Get the file's record."""
         return self.record_cls(self._record.data, model=self._record)
 
+    @classmethod
+    def remove_many(cls, record_file_ids):
+        """Hard delete record file instances."""
+        record_files = cls.model_cls.query.filter(cls.model_cls.id.in_(record_file_ids))
```
kpsherva, line 106: # for reviewers (important): 
I would like to get your feedback on this, it is a bulk removal so quite delicate operation, equivalent of previously used https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py#L506 (we used this method for removing file record instances one by one.
the problem is that in that method we have also implemented signals (pre and post delete), do you see any possible consequences of not having signals for file record? I might have not thought about everything

slint: We basically "deprecated" record signals for InvenioRDM, so I don't see any issues.

invenio_records_resources/records/systemfields/files/manager.py
```diff
@@ -249,6 +249,32 @@ def delete_all(self, remove_obj=True, softdelete_obj=True, remove_rf=False):
...
+        Deletes FileRecords, Bucket and ObjectVersions attached to file manager.
+        equivalent of:
+        softdelete_obj = false, remove_rf=True, remove_obj=true
+        option of deleting only FileRecords if full=False
+        """
+        file_record_ids = [entry.id for entry in self.entries.values()]
+        self.file_cls.remove_many(file_record_ids)
+
+        if full:
+            self.full_teardown()
```
slint, line 273: minor/nit: since `full_teardown` is only called here, I would just inline it to keep the logic all in one place.

slint (approved):
LGTM, one nit on inlining a method.

invenio_records_resources/records/systemfields/files/manager.py
```diff
@@ -249,6 +249,32 @@ def delete_all(self, remove_obj=True, softdelete_obj=True, remove_rf=False):
...
+
+    def teardown(self, full=True):
+        """Clean up all file manager related instances.
+
+        Deletes FileRecords, Bucket and ObjectVersions attached to file manager.
+        equivalent of:
+        softdelete_obj = false, remove_rf=True, remove_obj=true
+        option of deleting only FileRecords if full=False
+        """
+        file_record_ids = [entry.id for entry in self.entries.values()]
```
lnielsen, line 269: Comment: Will this not execute a query to get all files records? I would perhaps rather go with something like:

```python
self.file_cls.remove_all(record_id)
```

All file records should have a FK to the associated record.

invenio_records_resources/records/systemfields/files/manager.py
```diff
@@ -249,6 +249,32 @@ def delete_all(self, remove_obj=True, softdelete_obj=True, remove_rf=False):
                 remove_rf=remove_rf,
             )
 
+    def full_teardown(self):
+        """Hard delete bucket and Object versions associated with file manager."""
+        if self.enabled:
+            object_version_ids = [
+                entry.object_version_id for entry in self.entries.values()
+            ]
+            ObjectVersion.remove_many(object_version_ids, self.bucket)
```
lnielsen, line 258: Here as well, I would go with just a delete query that removes everything in the bucket.

slint (approved):
Nice, I think we got it this time :)

PR #542, opened 2023-11-03
manager: improe copy func

jrcastro2:
* closes https://github.com/zenodo/rdm-project/issues/508

requires https://github.com/inveniosoftware/invenio-files-rest/pull/299/files


tests/records/test_systemfield_files.py
```diff
@@ -384,7 +384,7 @@ def test_record_files_dump(base_app, db, location):
         "entries": [
             {
                 "uuid": str(rf.id),
-                "version_id": 3,
+                "version_id": 2,
```
kpsherva, line 387: to revert

invenio_records_resources/records/models.py
```diff
@@ -53,7 +53,7 @@ def object_version_id(cls):
     @declared_attr
     def object_version(cls):
         """Object version connected to the record file."""
-        return db.relationship(ObjectVersion)
+        return db.relationship(ObjectVersion)  # Can we cache this?
```
kpsherva, line 56: leftover comment?

invenio_records_resources/records/api.py
```diff
@@ -66,13 +66,14 @@ def get_by_key(cls, record_id, key):
     @classmethod
     def list_by_record(cls, record_id, with_deleted=False):
         """List all record files by record ID."""
-        query = cls.model_cls.query.filter(cls.model_cls.record_id == record_id)
+        with db.session.no_autoflush:
```
kpsherva, line 69: I am not sure about this change, could you explain how does it help?

invenio_records_resources/records/systemfields/files/manager.py
```diff
@@ -266,24 +270,61 @@ def teardown(self, full=True):
...
-            else:
-                dst_obj = rf.object_version
-
-            # Copy file record
-            if rf.metadata is not None:
-                self[key] = dst_obj, rf.metadata
-            else:
-                self[key] = dst_obj
+        bucket_objects = ObjectVersion.query.filter_by(bucket_id=self.bucket_id).count()
+        if bucket_objects < 1:
```
kpsherva, line 284:
```suggestion
        if bucket_objects == 0:
```
i would be worried if we have -1

invenio_records_resources/records/systemfields/files/manager.py
```diff
@@ -266,24 +270,61 @@ def teardown(self, full=True):
...
+
+            if rf_to_bulk_insert:
+                db.session.execute(insert(self.file_cls.model_cls), rf_to_bulk_insert)
+                # we need to populate entries from DB so we store the record file model
+                # instance
+                if not self._entries:
+                    self._entries = {}
+                    for rf in self.file_cls.list_by_record(self.record.id):
+                        self._entries[rf.key] = rf
+        else:
```
kpsherva, line 313: could we split this if statement to separate two methods instead? we know that the above block is used only while editing already published record, so we can use it directly

zzacharo: I agree, but the point was not to change the existing chain of calls i.e draft_files.copy(record_files). I can do it if we align with @slint as we had discussed IRL and we said that this could be done later.

invenio_records_resources/records/api.py
zzacharo: I think it was leftover from @jrcastro2  but in principle this is the fix we apply in other places instructing sqlalchemy not to trigger UPDATE queries when fetching from db. Is that the question?

jrcastro2: It helps very little, concretely it prevents 2 more queries to the DB every time this is called, the `no_autoflush` is useful when we are only querying, without executing any modifications on the data. Otherwise it can lead to inconsistent data. We can remove it if it's too complicated as the real gains are very little in this case.

setup.cfg
```diff
@@ -54,7 +54,7 @@ tests =
     invenio-db[postgresql,mysql,versioning]>=1.0.14,<2.0.0
     pytest-invenio>=2.1.0,<3.0.0
     pytest-mock>=1.6.0
-    sphinx>=4.2.0,<5
+    sphinx>=5,<6
```
jrcastro2, line 57: I had to update this in order for the tests to pass, the error was caused by one of the extensions
Screenshot from 2024-01-15 15-07-05
Apparently it was introduced some time ago but there was a release recently that maybe it's being picked up now, see here [changelog](https://github.com/sphinx-doc/sphinxcontrib-applehelp/blob/master/CHANGES#L22C47-L22C47).
The other alternative is to pin the extensions to <1.0.5

slint (approved):

PR #544, opened 2023-11-10
api: get file extension from key

jrcastro2:
* closes https://github.com/zenodo/rdm-project/issues/547

:warning: IMPORTANT: For this change to have effect it requires to reindex all records as the file types are indexed in OS

### DIFF
--- a/invenio_records_resources/records/api.py
+++ b/invenio_records_resources/records/api.py
@@ -10,6 +10,7 @@
 """Records API."""
 
 import mimetypes
+import os
 from contextlib import contextmanager
 
 from invenio_db import db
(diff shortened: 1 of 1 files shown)


invenio_records_resources/records/api.py
```diff
@@ -177,7 +178,9 @@ def dumps(self):
     @property
     def ext(self):
         """File extension."""
-        ext = mimetypes.guess_extension(self.object_model.mimetype)
```
lnielsen, line 180:
```python
# The ``ext`` property is used to in search to aggregate file types and we want e.g. both ``.jpeg`` and ``.jpg`` to be aggregated under ``.jpg``
ext = mimetypes.guess_extension(self.object_model.mimetype)
if ext is not None:
    return ext[1:]

# Support non-standard file extensions that cannot be guessed
_, ext = os.path.splitext(self.key)
if ext and len(ext) <= 5:
    return ext.lower()
```

jrcastro2: I see, thanks for the clarification.
However to make this work we would also need to change [this](https://github.com/inveniosoftware/invenio-files-rest/blob/master/invenio_files_rest/utils.py#L57) to not return "application/octet-stream". I will do the change

jrcastro2: Decided not to remove the fallback mimetype as it might cause issues on files serving, instead we check if the mimetype is set to the fallback to try to get the extenstion from the filename or not.

invenio_records_resources/records/api.py
```diff
@@ -10,10 +10,12 @@
 """Records API."""
 
 import mimetypes
+import os
 from contextlib import contextmanager
 
 from invenio_db import db
 from invenio_files_rest.models import FileInstance, ObjectVersion
+from invenio_files_rest.utils import FALLBACK_MIMETYPE
```
kpsherva, line 18: this import is very prone to compatibility issues - I find it a bit risky to import variables from utils of another module (normally I would expect it to be exposed as config, not hidden in utils, if it is supposed to be used by other modules)

PR #549, opened 2023-11-27
facets: provide CombinedTermsFacet to fix #798 [+]

fenekku:
This approach doesn't use 'nested' fields, but instead relies on a
field containing `<parent><split char><child>` to aggregate correctly.

- Alternative to: https://github.com/inveniosoftware/invenio-records-resources/pull/546
- Part of closing: https://github.com/inveniosoftware/invenio-rdm-records/issues/798

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -129,3 +129,6 @@ dmypy.json
 # Pyre type checker
 .pyre/
 .DS_Store
+
+# VSCode editor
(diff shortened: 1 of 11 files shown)


lnielsen (approved):

fenekku:
*the mergening has started*

PR #552, opened 2023-12-14
Changed default search type for SuggestQueryParser queries

monotasker:
### Description

Changes the default OpenSearch search type for the SuggestQueryParser. The current default search type (for the "multi_match" search) is "bool_prefix". This produces unexpected results because it does not differentiate between results based on weighted score. It also does not prioritize, e.g., matches near the beginning of a field or exact matches. I have changed this to "best_fields" which uses a weighted score to prioritize the best matches to the provided query string. The resulting hits are closer to a typical user's expectation. 

Because this is set as a default it can still be overridden when the SuggestQueryParser class is instantiated by passing a different "type" value in the "extra_params" keyword argument.

### Checklist

- [X] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [X] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [X] I've added relevant test cases.
- [X] I've added relevant documentation.
- [X] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/best-practices/i18n/) (for relevant code).
- [X] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines (for relevant code).
- [X] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines (for relevant code).
- [X] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines (for relevant code).
- [X] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [X] I've NOT included third-party code (copy/pasted source code or new dependencies).

**Third-party code**

If you've added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect](https://github.com/orgs/inveniosoftware/teams/architects/members).

**Reminder**

### DIFF
--- a/invenio_records_resources/services/records/queryparser/suggest.py
+++ b/invenio_records_resources/services/records/queryparser/suggest.py
@@ -51,7 +51,7 @@ class SuggestQueryParser(QueryParser):
     def __init__(self, identity=None, extra_params=None, **kwargs):
         """Constructor."""
         super().__init__(identity=identity, extra_params=extra_params)
-        self.extra_params.setdefault("type", "bool_prefix")
+        self.extra_params.setdefault("type", "best_fields")
 
     def parse(self, query_str):
         """Parse the query."""


invenio_records_resources/services/records/queryparser/suggest.py
```diff
@@ -51,7 +51,7 @@ class MySearchOptions(SearchOptions):
     def __init__(self, identity=None, extra_params=None, **kwargs):
         """Constructor."""
         super().__init__(identity=identity, extra_params=extra_params)
-        self.extra_params.setdefault("type", "bool_prefix")
+        self.extra_params.setdefault("type", "best_fields")
```
fenekku, line 54: So I was curious to understand better why this would give better results :thinking: ... I mean if you've observed it, I trust it. It's wasn't very clear to me *why* though.

From reading https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html it seems like `bool_prefix` does combine scores too (it sums the score across each field it targets). `best_fields` picks the highest score among the fields it targets (by assigning a higher score to a field, the more it contains the search terms). 

However, if I understand correctly it's the `prefix` part/query (well, its absence now) that makes a difference here; now, finding the term prefixing something in the field doesn't score higher. The `match` query used by `best_fields` under the hood (as opposed to multi `term` + `prefix`) probably gives a higher score to fields that have higher ratios of total characters to matched characters. So something like:

```
Search: "fr"
Over: "French", "Cajun French", "Old French (842-ca. 1400)", "Eastern Frisian"
```

would presumably score "French" higher now because it matches a higher ratio of the content...

Is that right?

fenekku (approved):
I think I can see why this would do better, but I am interested in knowing if my understand is correct (see below). No blocker though. This is even potentially more performant.

invenio_records_resources/services/records/queryparser/suggest.py
monotasker: Yes, you're right. I was over-simplifying. The `bool_prefix` search does look at summed scores across the fields. In our case because the fields being searched are manipulations of the one short vocabulary term string (e.g.,  "subject" and the derived "subject._2gram" and "subject._3gram"), I don't think the scores are going to vary between the fields. 

There is also some scoring involved in the search in individual fields, but it doesn't seem to be very meaningful for us. A "bool_prefix" search converts the last term in the search string to a prefix search. Prefix searches seem not to be weighted. They score a 1 if any word in the field begins with the prefix string, regardless of where that word comes in the field or (as you say) the proportion of the field it occupies. So in a vocabulary search, any search string that's a single term will produce a result set with identical scores. This is what I was finding in my initial tests with "bool_prefix": all of the results had a score of 100.

As you say, when the search string extends to more than one term, then only the last term is converted to a prefix search. The others are converted to exact term searches. Here there *is* a slight variation in scoring. But the only things that I've so far seen to affect the score are

1) whether the complete field text is a match to the exact term, and if not then
2) whether a longer field contains at least one exact match in a longer text, and
3) the number of exact matches in that longer text 

As you point out, where the whole field isn't an exact match, there don't appear to be any variation for, e.g., position of the match in the text or proportion of the text occupied by the match. There are likely factors I'm missing here, but the result is only a limited set of different scores that aren't that helpful. (The docs recommend against term searches for text fields.)  

For multi-term search strings, the bool_prefix search constructs a boolean search combining these prefix and term searches with the "should" operator. So a search for "computer programming" is converted to:
```
{
  "query": {
    "bool" : {
      "should": [
        { "term": { "subject": "computer" }},
        { "prefix": { "subject": "programming"}}
      ]
    }
  }
}
```
I haven't dug into how the "should" operator works here, and in my tests there seem to be some complexities that produce score variations. But they don't tend to be useful to us, and are in some cases counter-intuitive. 

The upshot is what you've said: that the factors we intuitively expect to be used--things like proportion of the text matched and position of the match in the text--aren't taken into account by "bool_prefix". And most problematic, there's *no* variation in scoring at all for single term queries.

And (again, as you point out) the "match" queries fit those intuitive expectations much better.

monotasker: I may have responded at the wrong level of the thread! Hopefully you can see that I responded, though. Thanks for approving the change.

fenekku: Thanks so much for that. I think that makes sense to me. Elastic/Open search is a beast. Someone else at CERN (or other collaborator) should give it a last check before we merge it, but that all sounds good to me. I am glad we have some general sense of what is going.

monotasker:
I notice one of the tests (test_search_suggest) failed because the test record wasn't suggested. I'll take a look and see whether that meets expectations given the change.

monotasker:
It turns out there is a crucial limitation to the "best_fields" query type: it doesn't match partial strings. This was obscured by the fact that we were getting matches to the 2/3_ngrams fields. But I'm no longer convinced this is a good solution. Closing the PR.

PR #554, opened 2024-01-18
models: add index on object_version_id

jrcastro2:
* closes https://github.com/zenodo/rdm-project/issues/640


invenio_records_resources/records/models.py
```diff
@@ -48,6 +48,7 @@ def object_version_id(cls):
             UUIDType,
             db.ForeignKey(ObjectVersion.version_id, ondelete="RESTRICT"),
             nullable=True,
+            index=True,
```
jrcastro2, line 51: This repo doesn't require an alembic recipe because this is just a mixin.

slint (approved):

PR #561, opened 2024-02-15
schema: fix data reference update in pre_load

yashlamba:
There seems to an issue with marshmallow that they are not freezing the original data which causes potential modifications leading to an updated copy of the object on subsequent pre/validate/post calls.

https://github.com/marshmallow-code/marshmallow/blob/13a12ad2eed262933770c643febbc46d4905c4c9/src/marshmallow/schema.py#L851-L853

This is work around for that for this specific case https://github.com/inveniosoftware/invenio-communities/pull/1088 where we need to use `pass_original`.

### DIFF
--- a/invenio_records_resources/services/records/schema.py
+++ b/invenio_records_resources/services/records/schema.py
@@ -8,7 +8,7 @@
 # details.
 
 """Record schema."""
-
+from copy import deepcopy
 from datetime import timezone
 
 from marshmallow import Schema, ValidationError, fields, pre_load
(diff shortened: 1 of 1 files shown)


jrcastro2 (approved):
LGTM! Thanks a lot for this fix :rocket:  I think it would be good if we can fix it in all other places that manipulate the data using the `pre_load`, such as  [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/00da0e44d61bddd237adccc70c08de66302a36c8/invenio_rdm_records/services/schemas/parent/__init__.py#L55-L74)

I would check in [all places](https://github.com/search?q=org%3Ainveniosoftware+%40pre_load&type=code) and apply the same fix so that we don't waste time in the future with this issue :wink:

invenio_records_resources/services/records/schema.py
```diff
@@ -36,6 +36,7 @@ def clean(self, data, **kwargs):
         Why: We want to allow the output of a Schema dump, to be a valid input
              to a Schema load without causing strange issues.
         """
+        data = deepcopy(data)
```
zzacharo, line 39: minor: maybe a rename 
```suggestion
        data = deepcopy(original_data)
```

yashlamba: `original_data` might not be the best since we can do `pass_original` on `pre_load`. Changed it to input_data.

PR #565, opened 2024-03-15
records: added access field to files

alejandromumo:
closes https://github.com/zenodo/zenodo-rdm/issues/777

* Added access field to files
* Added tests for access field
* Added explicit metadata field to files
* Updated tests for explicit metadata field
* Added schema loading for files update


slint:
Included in #574

PR #569, opened 2024-03-21
services: allow empty files to be uploaded

alejandromumo:
* Added config option to allow empty files to be uploaded.
* Added tests for empty files.


slint (approved):
Nice! LGTM, but I think we're in the weeds with the current state of CI

PR #573, opened 2024-04-04
records: add calculated system field with indexing

lnielsen:
* Moved from invenio-users-resources.

* Allows calculated system field to cache the value in the index, and use the cached value when reading from the index.

closes https://github.com/zenodo/ops/issues/398

### DIFF
--- a/invenio_records_resources/records/systemfields/calculated.py
+++ b/invenio_records_resources/records/systemfields/calculated.py
@@ -55,3 +55,24 @@ class CalculatedField(SystemField, abc.ABC):
     def calculate(self, record):
         """Logic for calculating the record's property."""
         return None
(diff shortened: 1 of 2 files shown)


invenio_records_resources/records/systemfields/calculated.py
```diff
@@ -55,3 +55,24 @@ def __set__(self, record, value):
     def calculate(self, record):
         """Logic for calculating the record's property."""
         return None
+
+
+class CalculatedIndexedField(CalculatedField):
+    """Field that also indexes it's calculated value."""
```
carlinmack, line 61:
```suggestion
    """Field that also indexes its calculated value."""
```

PR #575, opened 2024-05-14
errors: rework PermissionDeniedError

anikachurilova:
* closes https://github.com/inveniosoftware/product-rdm/issues/178

### DIFF
--- a/invenio_records_resources/services/errors.py
+++ b/invenio_records_resources/services/errors.py
@@ -15,6 +15,18 @@ from invenio_i18n import gettext as _
 from marshmallow import ValidationError
 
 
+class RecordPermissionDeniedError(PermissionDenied):
+    """Record permission denied error."""
+
+    description = "Permission denied."
+
+    def __init__(self, action_name=None, record=None, *args, **kwargs):
+        """Initialize exception."""
+        self.record = record
+        self.action_name = action_name
+        super(RecordPermissionDeniedError, self).__init__(*args, **kwargs)
+
+
 class PermissionDeniedError(PermissionDenied):
     """Permission denied error."""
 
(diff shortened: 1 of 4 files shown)


invenio_records_resources/services/base/service.py
```diff
@@ -48,7 +48,7 @@ def require_permission(self, identity, action_name, **kwargs):
         Like `check_permission` but raises an error if not allowed.
         """
         if not self.check_permission(identity, action_name, **kwargs):
-            raise PermissionDeniedError(action_name)
+            raise PermissionDeniedError(action_name, kwargs["record"])
```
anikachurilova, line 51: Needed it for this https://github.com/inveniosoftware/invenio-app-rdm/pull/2670/files#diff-2fb602eed821cbd7d07f7c61560b3e8ed50e165259da085c113fff47274b8496R426

invenio_records_resources/services/errors.py
```diff
@@ -20,6 +20,12 @@ class PermissionDeniedError(PermissionDenied):
 
     description = "Permission denied."
 
+    def __init__(self, action_name=None, record=None, *args, **kwargs):
```
kpsherva, line 23: due to modification done to the constructor signature this is breaking change 🔴 
are we sure this is the only way of passing record context to the error handler?

invenio_records_resources/services/errors.py
```diff
@@ -20,6 +20,12 @@ class PermissionDeniedError(PermissionDenied):
...
+        """Initialize exception."""
+        self.record = record
+        self.action_name = action_name
+        super(PermissionDeniedError, self).__init__(*args, **kwargs)
```
kpsherva, line 27: connected with the previous comment: is this parent constructor call matching parent class constructor signature? should it accept args and kwargs?

PR #576, opened 2024-05-19
processors: optionally use PyVIPS for image metadata

slint:
- If `pyvips` (and `libvips` on sysstem) are installed, PyVIPS is used to extract width and height for images.
- Compared to Wand, PyVIPS has the benefit of not loading the entire image in memory to extract width/height metadata for the image (works for multi-layer images also, such as TIFFs and PDFs). See [this gist](https://gist.github.com/slint/55c4ca8e1c59775e73942d4424a62c67) for a small memory analysis.

### DIFF
--- a/invenio_records_resources/services/files/processors/base.py
+++ b/invenio_records_resources/services/files/processors/base.py
@@ -8,7 +8,6 @@
 
 """Files processing engine."""
 
-
 import os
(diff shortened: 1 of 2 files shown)


yashlamba (approved):
LGTM!

PR #578, opened 2024-05-23
entity resolvers: resolve drafts for notification purposes

anikachurilova:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2667

### DIFF
--- a/invenio_records_resources/references/entity_resolvers/results.py
+++ b/invenio_records_resources/references/entity_resolvers/results.py
@@ -46,6 +46,7 @@ class ServiceResultResolver(EntityResolver):
         proxy_cls=ServiceResultProxy,
         item_cls=None,
         record_cls=None,
+        draft_cls=None,
     ):
         """Constructor.
 
@@ -58,6 +59,7 @@ class ServiceResultResolver(EntityResolver):
         self.proxy_cls = proxy_cls
         self._item_cls = item_cls
         self._record_cls = record_cls
+        self._draft_cls = draft_cls
 
     @property
     def item_cls(self):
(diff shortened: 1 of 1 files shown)


invenio_records_resources/references/entity_resolvers/results.py
```diff
@@ -69,9 +71,17 @@ def record_cls(self):
...
+    @property
+    def draft_cls(self):
+        """Get specified draft class or from service."""
+        return self._draft_cls or self.get_service().draft_cls
+
     def matches_entity(self, entity):
         """Check if the entity is a result item."""
-        return isinstance(entity, (self.item_cls, self.record_cls))
+        try:
+            return isinstance(entity, (self.item_cls, self.record_cls, self.draft_cls))
```
anikachurilova, line 82: Resolve draft to be able to send notifications for sharing access to drafts

invenio_records_resources/references/entity_resolvers/results.py
```diff
@@ -69,9 +71,17 @@ def record_cls(self):
...
+        except AttributeError:
```
kpsherva, line 83: I am a bit worried that attribute level might be a bit too generic and hide other exceptions (not only the ones produced by accessing self.item_cls self.record_cls, self.draft_cls), but I don't have a good alternative for now

invenio_records_resources/references/entity_resolvers/results.py
```diff
@@ -69,9 +71,21 @@ def record_cls(self):
...
+
     def matches_entity(self, entity):
         """Check if the entity is a result item."""
-        return isinstance(entity, (self.item_cls, self.record_cls))
+        is_item_or_record = isinstance(entity, (self.item_cls, self.record_cls))
+        if is_item_or_record:
+            return is_item_or_record
+
+        try:
+            return isinstance(entity, self.draft_cls)
```
kpsherva, line 86: after a round of a chat, I understood that this part should be probably placed in entity resolver dedicated to drafts

PR #579, opened 2024-05-30
uow: update TaskOp; add TaskRevokeOp

yashlamba:
closes https://github.com/inveniosoftware/invenio-jobs/issues/6 (edited  by @alejandromumo  🥷)

Earlier `TaskOp` used the Celery task delay method, so the signature of init was with respect to delay. However, now we also want to pass execution options which are only possible through `apply_async` (Ref for the signature difference: https://docs.celeryq.dev/en/stable/userguide/calling.html#example).

The `for_apply_async` receives `apply_async` compatible arguments and returns a valid `TaskOp` while using `TaskOp` directly will support the existing signature of `delay`, for backwards compatibility.

### DIFF
--- a/invenio_records_resources/services/uow.py
+++ b/invenio_records_resources/services/uow.py
@@ -104,6 +104,7 @@ class and implementing the desired methods:
 
 from functools import wraps
 
+from celery import current_app
 from invenio_db import db
 
 from ..tasks import send_change_notifications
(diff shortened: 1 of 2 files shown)


invenio_records_resources/services/uow.py
```diff
@@ -261,10 +262,32 @@ def __init__(self, celery_task, *args, **kwargs):
...
+class TaskRevokeOp(Operation):
+    """A celery task stopping operation."""
+
+    def __init__(self, task_id: str) -> None:
+        """Initialize the task operation."""
+        self.task_id = task_id
+
+    def on_post_commit(self, uow) -> None:
+        """Run the revoke post commit."""
+        current_app.control.revoke(self.task_id, terminate=True)
```
ntarocco, line 290: Shall we add a check if `.control` exist, basically if current_app is a Celery app first?

invenio_records_resources/services/uow.py
```diff
@@ -261,10 +262,32 @@ def __init__(self, celery_task, *args, **kwargs):
...
 
     def on_post_commit(self, uow):
         """Run the post task operation."""
-        self._celery_task.delay(*self._args, **self._kwargs)
+        self._celery_task.apply_async(
+            args=self._args, kwargs=self._kwargs, **self.celery_kwargs
+        )
+
+    @classmethod
+    def for_async_apply(cls, celery_task, args=None, kwargs=None, **celery_kwargs):
```
ntarocco, line 274: Quick question: is this added just as a helper/util to set the defaults args to `tuple`?
I think we call this `def create` in another places, unless I am misunderstanding the need ☺️

ntarocco (commented):
Looks good! Added a couple of small comments ☺️

invenio_records_resources/services/uow.py
yashlamba: do you have a recommendation of what to do if it's not? Raise or pass?

alejandromumo: I believe you can get the current celery app from the `invenio-celery` [extension](https://github.com/inveniosoftware/invenio-celery/blob/master/invenio_celery/ext.py)

alejandromumo (commented):
I left a small comment, otherwise LGTM.

PR #581, opened 2024-07-02
files: sync metadata on file edit

yashlamba:
We calculate image metadata when files are committed, which can happen on record edit as well. 

Right now we only sync the file object version but not the metadata on edit - this PR fixes that.

### DIFF
--- a/invenio_records_resources/records/systemfields/files/manager.py
+++ b/invenio_records_resources/records/systemfields/files/manager.py
@@ -360,7 +360,12 @@ class FilesManager(MutableMapping):
                 if obj_or_key in self:
                     del self[obj_or_key]
             elif operation == "add":
-                self[obj_or_key.key] = obj_or_key
+                f_key = obj_or_key.key
+                rf = src_files[f_key]
+                if rf.metadata is not None:
+                    self[f_key] = obj_or_key, dict(rf)
+                else:
+                    self[f_key] = obj_or_key
 
     @property
     def entries(self):


invenio_records_resources/records/systemfields/files/manager.py
```diff
@@ -360,7 +360,7 @@ def sync(self, src_files, delete_extras=True):
                 if obj_or_key in self:
                     del self[obj_or_key]
             elif operation == "add":
-                self[obj_or_key.key] = obj_or_key
+                self[obj_or_key.key] = obj_or_key, dict(src_files[obj_or_key.key])
```
kpsherva, line 363: just to understand what is happening: why do you need to cast it to dict? what type does `src_files[obj_or_key.key]` return ?

yashlamba: I updated the code a bit more to understand, essentially, I am trying to replicate exactly what's happening here: https://github.com/inveniosoftware/invenio-records-resources/blob/47bf68f6eaca88f004f7f642827552fc77edf021/invenio_records_resources/records/systemfields/files/manager.py#L324-L328

`src_files[obj_or_key.key]` <- gives the latest file metadata

invenio_records_resources/records/systemfields/files/manager.py
```diff
@@ -360,7 +360,12 @@ def sync(self, src_files, delete_extras=True):
                 if obj_or_key in self:
                     del self[obj_or_key]
             elif operation == "add":
-                self[obj_or_key.key] = obj_or_key
+                f_key = obj_or_key.key
+                rf = src_files[f_key]
+                if rf.metadata is not None:
+                    self[f_key] = obj_or_key, dict(rf)
```
alejandromumo, line 366: `dict(rf)` returns the metadata only?

alejandromumo (approved):

zzacharo (approved):

PR #585, opened 2024-07-16
service: add create update many

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-vocabularies/issues/353
* NEEDS https://github.com/inveniosoftware/invenio-records-permissions/pull/100

### DIFF
--- a/invenio_records_resources/services/records/service.py
+++ b/invenio_records_resources/services/records/service.py
@@ -591,63 +591,73 @@ class RecordService(Service, RecordIndexerMixin):
         :param identity: The user identity performing the operation.
         :param data: A list of tuples containing the record ID and record data.
         :param uow: The unit of work to register the record operations. Defaults to None.
-
-        Returns:
-            list: A list of tuples containing the operation type ('create' or 'update'), the processed record or the record dict, and any schema errors encountered.
         """
         records_processed = []
+
+        self.require_permission(identity, "create_or_update_many")
+
+        def _update_record(record_dict, record):
+            """Update a record."""
+            record_data, schema_errors = self.schema.load(
+                record_dict,
+                context=dict(identity=identity, pid=record.pid, record=record),
+                raise_errors=False,
+            )
+
+            # If errors we avoid creating the record
+            if schema_errors:
+                records_processed.append(("update", record_dict, schema_errors, None))
+                return
+
+            # Run components
+            self.run_components(
+                "update", identity, data=record_data, record=record, uow=uow
+            )
+
+            records_processed.append(("update", record, schema_errors, None))
+
+        def _create_record(record_dict):
+            """Create a record."""
+            # Validate data and create record with pid
+            record_data, schema_errors = self.schema.load(
+                record_dict, context={"identity": identity}, raise_errors=False
+            )
+
+            # If errors we avoid creating the record
+            if schema_errors:
+                records_processed.append(("create", record_dict, schema_errors, None))
+                return
+
+            # It's the components who saves the actual data in the record
+            record = self.record_cls.create({})
+
+            # Run components
+            self.run_components(
+                "create",
+                identity,
+                data=record_data,
+                record=record,
+                errors=schema_errors,
+                uow=uow,
+            )
+            records_processed.append(("create", record, schema_errors, None))
+
+        # We avoid using create and update methods to bulk index all records at once
         for record_id, record_dict in data:
             try:
                 record = self.record_cls.pid.resolve(record_id)
-
-                # Permissions
-                self.require_permission(identity, "update", record=record)
-                record_data, schema_errors = self.schema.load(
-                    record_dict,
-                    context=dict(identity=identity, pid=record.pid, record=record),
-                    raise_errors=False,
-                )
-
-                # If errors we avoid creating the record
-                if schema_errors:
-                    records_processed.append(("update", record_dict, schema_errors))
-                    continue
-
-                # Run components
-                self.run_components(
-                    "update", identity, data=record_data, record=record, uow=uow
-                )
-
-                records_processed.append(("update", record, schema_errors))
-            except PIDDoesNotExistError:
-                self.require_permission(identity, "create")
-
-                # Validate data and create record with pid
-                record_data, schema_errors = self.schema.load(
-                    record_dict, context={"identity": identity}, raise_errors=False
-                )
-
-                # If errors we avoid creating the record
-                if schema_errors:
-                    records_processed.append(("create", record_dict, schema_errors))
-                    continue
-
-                # It's the components who saves the actual data in the record.
-                record = self.record_cls.create({})
-
-                # Run components
-                self.run_components(
-                    "create",
-                    identity,
-                    data=record_data,
-                    record=record,
-                    errors=schema_errors,
-                    uow=uow,
-                )
-                records_processed.append(("create", record, schema_errors))
+                _update_record(record_dict, record)
+            except (NoResultFound, PIDDoesNotExistError):
+                _create_record(record_dict)
+            except Exception as exc:
+                records_processed.append(("create", record_dict, None, exc))
 
         # We only commit records that have no errors
-        records = [record for _, record, errors in records_processed if errors == []]
+        records = [
+            record
+            for _, record, errors, exc in records_processed
+            if errors == [] and exc is None
+        ]
         uow.register(RecordBulkCommitOp(records, self.indexer))
 
-        return records_processed
+        return self.result_bulk_list(self, identity, records_processed)
--- a/invenio_records_resources/services/records/service.py
+++ b/invenio_records_resources/services/records/service.py
@@ -17,6 +17,8 @@ from invenio_records_permissions.api import permission_filter
 from invenio_search import current_search_client
 from invenio_search.engine import dsl
 from kombu import Queue
+from marshmallow import ValidationError
+from sqlalchemy.orm.exc import NoResultFound
 from werkzeug.local import LocalProxy
 
 from invenio_records_resources.services.errors import (
@@ -26,7 +28,7 @@ from invenio_records_resources.services.errors import (
 
 from ..base import LinksTemplate, Service
 from ..errors import RevisionIdMismatchError
-from ..uow import RecordCommitOp, RecordDeleteOp, unit_of_work
+from ..uow import RecordBulkCommitOp, RecordCommitOp, RecordDeleteOp, unit_of_work
 from .schema import ServiceSchemaWrapper
 
 
@@ -579,3 +581,73 @@ class RecordService(Service, RecordIndexerMixin):
 
             self.reindex(identity, search_query=search_query)
         return True
+
+    @unit_of_work()
+    def create_or_update_many(self, identity, data, uow=None):
+        """Create or update a list of records.
+
+        This method takes a list of record data and creates or updates the corresponding records.
+
+        :param identity: The user identity performing the operation.
+        :param data: A list of tuples containing the record ID and record data.
+        :param uow: The unit of work to register the record operations. Defaults to None.
+
+        Returns:
+            list: A list of tuples containing the operation type ('create' or 'update'), the processed record or the record dict, and any schema errors encountered.
+        """
+        records_processed = []
+        for record_id, record_dict in data:
+            try:
+                record = self.record_cls.pid.resolve(record_id)
+
+                # Permissions
+                self.require_permission(identity, "update", record=record)
+                record_data, schema_errors = self.schema.load(
+                    record_dict,
+                    context=dict(identity=identity, pid=record.pid, record=record),
+                    raise_errors=False,
+                )
+
+                # If errors we avoid creating the record
+                if schema_errors:
+                    records_processed.append(("update", record_dict, schema_errors))
+                    continue
+
+                # Run components
+                self.run_components(
+                    "update", identity, data=record_data, record=record, uow=uow
+                )
+
+                records_processed.append(("update", record, schema_errors))
+            except PIDDoesNotExistError:
+                self.require_permission(identity, "create")
+
+                # Validate data and create record with pid
+                record_data, schema_errors = self.schema.load(
+                    record_dict, context={"identity": identity}, raise_errors=False
+                )
+
+                # If errors we avoid creating the record
+                if schema_errors:
+                    records_processed.append(("create", record_dict, schema_errors))
+                    continue
+
+                # It's the components who saves the actual data in the record.
+                record = self.record_cls.create({})
+
+                # Run components
+                self.run_components(
+                    "create",
+                    identity,
+                    data=record_data,
+                    record=record,
+                    errors=schema_errors,
+                    uow=uow,
+                )
+                records_processed.append(("create", record, schema_errors))
+
+        # We only commit records that have no errors
+        records = [record for _, record, errors in records_processed if errors == []]
+        uow.register(RecordBulkCommitOp(records, self.indexer))
+
+        return records_processed
--- a/invenio_records_resources/services/uow.py
+++ b/invenio_records_resources/services/uow.py
@@ -183,6 +183,27 @@ class RecordCommitOp(Operation):
             self._indexer.index(self._record, arguments=arguments)
 
 
+class RecordBulkCommitOp(Operation):
+    """Record bulk commit operation with indexing."""
+
+    def __init__(self, records, indexer=None, index_refresh=False):
+        """Initialize the bulk record commit operation."""
+        self._records = records
+        self._indexer = indexer
+        self._index_refresh = index_refresh
+
+    def on_register(self, uow):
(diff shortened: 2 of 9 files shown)


invenio_records_resources/services/records/service.py
```diff
@@ -579,3 +581,65 @@ def on_relation_update(
 
             self.reindex(identity, search_query=search_query)
         return True
+
+
+    @unit_of_work()
+    def create_update_many(self, identity, data, revision_id=None, uow=None, expand=False):
```
yashlamba, line 587: rename: create_or_update_many

invenio_records_resources/services/records/service.py
```diff
@@ -579,3 +581,65 @@ def on_relation_update(
...
+                records_to_update.append(record)
+            except NoResultFound:
+                try:
+                    self.require_permission(identity, "create")
+                    
+                    # Validate data and create record with pid
+                    record_data, schema_errors = self.schema.load(
+                        record_data,
+                        context={"identity": identity}
+                    )
```
yashlamba, line 621: check whether this raises errors by default

invenio_records_resources/services/records/service.py
```diff
@@ -579,3 +581,65 @@ def on_relation_update(
...
+                    errors=schema_errors,
+                    uow=uow,
+                )
+                
+                records_to_create.append(record)
+
+        records = records_to_update + records_to_create
+        uow.register(RecordBulkCommitOp(records, self.indexer))
+
+        return records, errors
```
yashlamba, line 645:
```python
return [
    (OPERATION_TYPE(created/updated), record_data, errors),
    ... 
]
```

invenio_records_resources/services/uow.py
```diff
@@ -182,6 +182,29 @@ def on_commit(self, uow):
...
+        self._indexer = indexer
+        self._index_refresh = index_refresh
+        
+
+    def on_register(self, uow):
+        """Save objects to the session."""
+        for record in self._records:
+            json = record._validate()
+            record.model.json = json
+            uow.session.merge(record.model)
```
yashlamba, line 200:
```suggestion
        for record in self._records:
            record.commit()
```

yashlamba (commented):
As discussed IRL

invenio_records_resources/services/records/service.py
```diff
@@ -579,3 +581,74 @@ def on_relation_update(
...
+                self.require_permission(identity, "update", record=record)
+                record_data, schema_errors = self.schema.load(
+                    record_dict,
+                    context=dict(identity=identity, pid=record.pid, record=record),
+                    raise_errors=False,
+                )
+
+                # If errors we avoid creating the record
+                if schema_errors:
+                    records_processed.append(("update", record_dict, schema_errors))
```
jrcastro2, line 614:
```suggestion
                    record = self.recordDTO(record_dict)
                    records_processed.append(("update", record_dict, schema_errors))
```

When it fails we are returning the record dict but on success we return the record class, should we try to wrap in a DTO class to be consistent with what we return? or are we good with this?

Actually this might be a bigger change to implement, just sharing my thoughts in case you guys have better ideas for this.

utnapischtim: my 2 cents: as someone using those methods, i would prefer to be consistent what will be returned. it is already irritating that no RecordItem will be returned as from the other methods.

invenio_records_resources/services/uow.py
```diff
@@ -183,6 +183,27 @@ def on_commit(self, uow):
...
+
+    def __init__(self, records, indexer=None, index_refresh=False):
+        """Initialize the bulk record commit operation."""
+        self._records = records
+        self._indexer = indexer
+        self._index_refresh = index_refresh
+
+    def on_register(self, uow):
+        """Save objects to the session."""
+        for record in self._records:
```
ntarocco, line 197: do we really need this? not sure it is very useful, compared to each record.commit op

invenio_records_resources/services/records/service.py
```diff
@@ -579,3 +581,74 @@ def on_relation_update(
...
+        Args:
+            identity (object): The user identity performing the operation.
+            data (list): A list of tuples containing the record ID and record data.
+            uow (UnitOfWork, optional): The unit of work to register the record operations. Defaults to None.
+
+        Returns:
+            list: A list of tuples containing the operation type ('create' or 'update'), the processed record or the record dict, and any schema errors encountered.
+        """
+        records_processed = []
+        for record_id, record_dict in data:
```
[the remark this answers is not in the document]
jrcastro2: Hmm to use them we would need to change the logic and signature to be able to avoid saving the record in the DB when we have validation errors (and also pass the raise_errors=False), at the moment if there are errors it would either raise and stop or continue and save the record with the errors anyway. In the vocabs we only want to save the record if it succeed otherwise skip it and simply log it. 

If you consider that in the `RecordService` class the `create_or_update_many` should behave using the base `create` and `update` methods maybe we can simply override this method in the vocabs service. Let me know what you think we should do in this case.

invenio_records_resources/services/uow.py
jrcastro2: Not sure I fully understand this comment, sorry ... but the point of RecordBulkCommitOp is to be able to bulk_index everything at once, after commiting it ot the DB, alternatively we could register the RecordCommitOp for each record (without the indexer) and register at the end the `RecordBulkIndexOp` to bulk index everything but I thought having on Op that does it together would be clearer

invenio_records_resources/services/records/service.py
```diff
@@ -579,3 +581,73 @@ def on_relation_update(
...
         return True
+
+    @unit_of_work()
+    def create_or_update_many(self, identity, data, uow=None):
+        """Create or update a list of records.
+
+        This method takes a list of record data and creates or updates the corresponding records.
+
+        :param identity: The user identity performing the operation.
+        :param data: A list of tuples containing the record ID and record data.
```
kpsherva, line 592: just for consistency reasons: are we using tuples also in the member bulk operations? it would be good to have a uniform data structure for all the bulk operations in the application

invenio_records_resources/services/records/service.py
```diff
@@ -579,3 +581,73 @@ def on_relation_update(
...
+        :param uow: The unit of work to register the record operations. Defaults to None.
+
+        Returns:
+            list: A list of tuples containing the operation type ('create' or 'update'), the processed record or the record dict, and any schema errors encountered.
```
kpsherva, line 596: nit: the format of the docstring here suggest we use typing (list:) which is not the case

invenio_records_resources/services/records/service.py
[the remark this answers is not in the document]
kpsherva: how is it done for the bulk members operation? if I remember correctly we were also gathering the errors from bulk operation in some data structure (not saying this might be needed here!)
is the requirement to abort and not save any records on the list if one fails or we should save only the correct one and omit the failed ones?

invenio_records_resources/services/records/service.py
jrcastro2: In members seems that we simply return True after processing everything, see here: https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/members/services/service.py#L232

I got inspired here with what we did for the including multiples records in a community, where we return a tuple of processed records and the errors. 

In this case we changed it a bit because ideally we want to keep track of the error, to also have the record data and operation tupe that caused the error.

invenio_records_resources/services/records/service.py
jrcastro2: Unless I am missing something orn ot understanding the comment, we are in fact returning a list of tuples

invenio_records_resources/services/records/service.py
kpsherva: could you explain what DTO expands to? I am missing some context, and in general, it would be nice not to use abbreviations in the method names

invenio_records_resources/services/records/service.py
```diff
@@ -579,3 +581,73 @@ def on_relation_update(
...
+                    records_processed.append(("update", record_dict, schema_errors))
+                    continue
+
+                # Run components
+                self.run_components(
+                    "update", identity, data=record_data, record=record, uow=uow
+                )
+
+                records_processed.append(("update", record, schema_errors))
+            except (NoResultFound, PIDDoesNotExistError):
```
kpsherva, line 622: I am curious why sometimes we can get NoResultFound and PIDDoesNotExistError, does it mean that we mix the way of accessing the record?

invenio_records_resources/services/records/service.py
jrcastro2: Sure, DTO stands for Data Object Transfer. It would allow us to avoid returning some times a dict class of the record -when it fails - and a Record class - when it succeded. The DTO object would take care of how to access the data. However another simpler alternative is to simply return a dict class of the record all the time.

invenio_records_resources/services/records/service.py
jrcastro2: Right, not sure why I added `NoResultFound`, I will test it again but should rely only on `PIDDoesNotExistError`

invenio_records_resources/services/records/service.py
slint: I see that the result type for the new service call is a big discussion topic in this PR.

Should we create a new `BulkResult` class, with an appropriate `BulkResultItem` (or better named?) container type to wrap all the necessary fields/info in an organized way? The main issue I see is that we have two ways to represent errors from service calls in RDM:

- Populating the `errors` list and including it in the result (e.g. for schema loading validation errors, or errors in service components)
  - We're very inconsistent about what goes in this list though, sometimes it's a [Marshmallow-style dictionary](https://github.com/inveniosoftware/invenio-drafts-resources/blob/607597db3bcaff600462a05521d3635c9032d481/invenio_drafts_resources/services/records/components/base.py#L114-L121), and other times just [a string](https://github.com/slint/invenio-rdm-records/blob/0a14ba9dc06645420d7543275b56a01dc88ede07/invenio_rdm_records/services/components/access.py#L38-L40)
- Raising an exception so that the caller can handle it
  - I would prefer being more "transactional" with the result types, meaning that raised exceptions are just wrapped, and it's up to the caller to handle them depending on their context
  - But this also can be dangerous since if the caller doesn't check if the results have errors, they might miss it (while a raised exception "forces" you to handle it)

My preference for bulk operations would be not to raise but instead wrap everything.

invenio_records_resources/services/records/service.py
slint: Some vocabularies don't rely on `PersistentIdentifier` in which case they don't raise a `PIDDoesNotExistError`, but instead `NoResultFound` (since they do a direct query to the DB on the table that holds the vocabulary).

invenio_records_resources/services/records/service.py
```diff
@@ -579,3 +581,73 @@ def on_relation_update(
...
+                    record_dict, context={"identity": identity}, raise_errors=False
+                )
+
+                # If errors we avoid creating the record
+                if schema_errors:
+                    records_processed.append(("create", record_dict, schema_errors))
+                    continue
+
+                # It's the components who saves the actual data in the record.
+                record = self.record_cls.create({})
```
kpsherva, line 636: would it make sense to split create_many and update_many to have dedicated methods? it would help to avoid having too much responsibilities handled in one place

invenio_records_resources/services/records/service.py
kpsherva: I am suggesting to remove the name of the returned type, not sure if we have it in any other docstring

invenio_records_resources/services/records/service.py
kpsherva: I agree with the BulkResult solution and BulkResultItem potentially (isn't it just ResultItem? I think only the "collection" needs a new class)

invenio_records_resources/services/records/service.py
kpsherva: I would be very hesitant to write another method just to deal with inconsistency of what we return. It does not solve the problem, just hides it.

jrcastro2: Addressed with new `ServiceBulkListResult`

invenio_records_resources/services/records/service.py
jrcastro2: PR updated with the above mentioned changes

invenio_records_resources/services/records/service.py
jrcastro2: Good point! For sure there is room to improve it. I feel like it might be a bit more complicated than what we think because right now we are resolving the PID to know if a record has to be updated or created, then we should get a list with those and then pass the list of the data to 2 different funcs (update | create) + handle the exceptions there as well. I would say that we can log this as an improvement, if you agree.

slint (approved):

invenio_records_resources/services/records/service.py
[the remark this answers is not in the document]
slint: The reason is because the default Service methods for `create`/`update` do direct indexing, while we want to send the records for bulk indexing. I think we should add a comment explaining this here, just so that it's a bit more explicit for anyone reading this part.

invenio_records_resources/services/records/service.py
```diff
@@ -579,3 +581,85 @@ def on_relation_update(
...
+            if schema_errors:
+                records_processed.append(("update", record_dict, schema_errors, None))
+                return
+
+            # Run components
+            self.run_components(
+                "update", identity, data=record_data, record=record, uow=uow
+            )
+
+            records_processed.append(("update", record, schema_errors, None))
```
utnapischtim, line 617: is it on purpose that the "update" key doesn't match the function name `_create_record`. same for `def _update_record`

invenio_records_resources/services/records/service.py
```diff
@@ -579,3 +581,85 @@ def on_relation_update(
...
+            )
+            records_processed.append(("create", record, schema_errors, None))
+
+        # We avoid using create and update methods to bulk index all records at once
+        for record_id, record_dict in data:
+            try:
+                record = self.record_cls.pid.resolve(record_id)
+                _create_record(record_dict, record)
+            except (NoResultFound, PIDDoesNotExistError):
+                _update_record(record_dict)
```
slint, line 653:
```suggestion
                _update_record(record_dict, record)
            except (NoResultFound, PIDDoesNotExistError):
                _create_record(record_dict)
```

I think these should be swapped?

jrcastro2: .... Right! Thanks for realising!

invenio_records_resources/services/records/service.py
jrcastro2: Yes, the permissions are the same as for `update` and `create`. Ideally we would want to have one permission check at the `create_update_many` (also to be more efficient) but it's not easy at the moment as for update we need to pass the record and for create we don't. Something to improve for the future.

utnapischtim: for me it is counterintuitive that the `update` keyword is used in the in the `_create_record` function. the name of the function is `_create_record` the docstring says it creates records but the the `run_components` call uses `update`

when you have a look to the `def _create` function you see that the for the [permission call](https://github.com/inveniosoftware/invenio-records-resources/blob/cefb51669bf5aa98f92631c098a439092f3bb168/invenio_records_resources/services/records/service.py#L341) for the [run_components call](https://github.com/inveniosoftware/invenio-records-resources/blob/cefb51669bf5aa98f92631c098a439092f3bb168/invenio_records_resources/services/records/service.py#L356) `create` as keyword is used.

slint: For the permissions checks maybe it makes sense that we just do the one check in the beginning of the `create_or_update_many` service call for a `can_create_or_update_many`? It should only allow for the system identity.

invenio_records_resources/services/records/service.py
```diff
@@ -579,3 +581,74 @@ def on_relation_update(
...
+        Returns:
+            list: A list of tuples containing the operation type ('create' or 'update'), the processed record or the record dict, and any schema errors encountered.
+        """
+        records_processed = []
+        for record_id, record_dict in data:
+            try:
+                record = self.record_cls.pid.resolve(record_id)
+
+                # Permissions
+                self.require_permission(identity, "update", record=record)
```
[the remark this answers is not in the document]
slint: See https://github.com/inveniosoftware/invenio-records-resources/pull/585#discussion_r1698016979 where we might just move a single permission check in  the beginning of the service call

invenio_records_resources/services/records/service.py
jrcastro2: Done, added https://github.com/inveniosoftware/invenio-records-permissions/pull/100

slint: Merged and released: https://pypi.org/project/invenio-records-permissions/0.20.0/

PR #592, opened 2024-08-08
errors: add error handler for community submission closed

anikachurilova:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/158

### DIFF
--- a/invenio_records_resources/services/errors.py
+++ b/invenio_records_resources/services/errors.py
@@ -110,3 +110,9 @@ class FilesCountExceededException(Exception):
                 )
             )
         )
+
+
+class RecordSubmissionClosedCommunityError(PermissionDenied):
+    """Record submission policy forbids non-members from submitting records to community."""
+
+    description = "Submission to this community is only allowed to community members."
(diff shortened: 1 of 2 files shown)


invenio_records_resources/services/errors.py
```diff
@@ -110,3 +110,9 @@ def __init__(self, max_files, resulting_files_count):
                 )
             )
         )
+
+
+class RecordSubmissionClosedCommunityError(PermissionDenied):
+    """Record submission policy forbids non-members from submitting records to community."""
+
+    description = "Submission to this community is only allowed to community members."
```
slint, line 118: minor: `invenio-records-resources` is too low in the dependency chain to know  about communities. Shouldn't this be in a module further up the chain?

kpsherva: +1 for this, there shouldn't be any references to communities in the records-resources modules, since communities depends on records-resources, not the other way around

anikachurilova: Moved changes to https://github.com/inveniosoftware/invenio-rdm-records/pull/1790
closing this PR

PR #594, opened 2024-08-14
services: Add locale to search params via args

sakshamarora1:
Fixes: https://github.com/inveniosoftware/invenio-rdm-records/issues/1791

We are currently searching on all title.* fields which causes the search to fail due to the `too_many_nested_clauses` exception.

This change will enable the `SearchOptions` class and `QueryParser` class(es) to be able to dyanmically get the locale and dynamically use it for searching

### DIFF
--- a/invenio_records_resources/resources/records/args.py
+++ b/invenio_records_resources/resources/records/args.py
@@ -9,6 +9,8 @@
 
 """Schemas for parameter parsing."""
 
+from invenio_i18n import get_locale
+
 from flask_resources.parsers import MultiDictSchema
 from marshmallow import fields, post_load, validate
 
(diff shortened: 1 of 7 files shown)


slint:
I feel this is a bit "intrusive" to the search options API, in the sense that it depends on injecting and "prop-drilling" the `locale` value throughout calls. It's also a backwards-incompatible change, since the methods where `query=None` has been added, don't by default have `**kwargs` in their signatures.

Since `get_locale` is actually relying on the request/session/application global state, it could be that we can inject it as a lazy-loaded value in places. What I had in mind is something like this:


```python
from invenio_i18n import get_locale
localized_title = LocalProxy(lambda: f"title.{get_locale()}^20")


class AffiliationsSearchOptions(SearchOptions):
    """Search options."""

    suggest_parser_cls = SuggestQueryParser.factory(
        fields=["name^100", "acronym^50", localized_title, "id^20", "aliases^20"],
    )

    ...
```

We could also make a utility that injects the locale in format-style string, which for end usable would look like `_l("title.{locale}^20")`, and would handle the "laziness".

That way we could avoid the braking change. Let me know if that makes sense.

PR #595, opened 2024-09-16
Errors: Add CommunityNotSelectedError for conditional check on config

sakshamarora1:
closes: https://github.com/CERNDocumentServer/cds-rdm/issues/88

### DIFF
--- a/invenio_records_resources/resources/errors.py
+++ b/invenio_records_resources/resources/errors.py
@@ -40,6 +40,7 @@ from ..services.errors import (
     QuerystringValidationError,
     RecordPermissionDeniedError,
     RevisionIdMismatchError,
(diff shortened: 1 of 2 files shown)


invenio_records_resources/resources/errors.py
```diff
@@ -207,4 +209,16 @@ class ErrorHandlersMixin:
                 description="Uploading selected files will result in exceeding the max amount per record.",
             )
         ),
+        CommunityNotSelectedError: create_error_handler(
```
kpsherva, line 212: this module is not aware of communities (invenio-communities depends on invenio-records-resources, not the other way around) so this handler should be declared either in communities or in rdm-records (a module that is aware both of the record and communities is the correct place)

PR #597, opened 2024-10-08
model: make forward compatible to sqlalchemy >= 2

utnapischtim:


utnapischtim:
seems not necessary anymore to migrate to flask>=3.0. the problem was in pytest-invenio. the change to not use the flask-sqlalchemy session was the problem. and the change has been done, because of the change in flask-sqlalchemy to not use bind anymore

PR #599, opened 2024-11-01
queryparser: add RestrictedTerm to exclude allowed phrases by permission

kpsherva:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/240

### DIFF
--- a/invenio_records_resources/services/records/queryparser/transformer.py
+++ b/invenio_records_resources/services/records/queryparser/transformer.py
@@ -14,6 +14,7 @@ how to build your own query tree transformer.
 """
 
 from invenio_i18n import gettext as _
+from luqum.tree import Phrase, Word
 from luqum.visitor import TreeTransformer
 
 from invenio_records_resources.services.errors import QuerystringValidationError
@@ -33,13 +34,66 @@ class FieldValueMapper:
         """Get the term name."""
         return self._term_name
 
-    def map_word(self, node):
+    def map_word(self, node, **kwargs):
         """Modify a word node."""
         return self._word_fun(node) if self._word_fun else node
 
-    def map_phrase(self, node):
+    def map_phrase(self, node, **kwargs):
         """Modify a phrase node."""
         return self._phrase_fun(node) if self._phrase_fun else node
+
+
+class RestrictedTerm:
+    """Class used to apply specific permissions to search.
+
+    ex. "internal_notes.note": RestrictedTerm(system_access_permission)
+    restricts the internal_notes.note field from being searched by anyone else
+    than system process
+    """
+
+    def __init__(self, permission):
+        """Constructor."""
+        self.permission = permission
+
+    def allows(self, identity):
+        """Check the permission for the whole term."""
+        return self.permission.allows(identity)
+
+
+class RestrictedTermValue:
+    """Class used to apply specific permissions to search specific words.
+
+    ex. "_exists_": RestrictedTermValue(
+            system_access_permission, word=word_internal_notes
+        ),
+    will rewrite the searched field based on given permission
+    """
+
+    def __init__(self, permission, word=None, phrase=None):
+        """Constructor."""
+        self.permission = permission
+        self._word_fun = word
+        self._phrase_fun = phrase
+
+    def map_word(self, node, context, **kwargs):
+        """Modify a word node."""
+        is_restricted = not self.permission.allows(context["identity"])
+        if not is_restricted:
+            return node
+        if self._word_fun and is_restricted:
+            return self._word_fun(node)
+        else:
+            return Word("")
+
+    def map_phrase(self, node, context, **kwargs):
+        """Modify a phrase node."""
+        is_restricted = not self.permission.allows(context["identity"])
+        if not is_restricted:
+            return node
+        if self._phrase_fun and is_restricted:
+            return self._phrase_fun(node)
+        else:
+            return Phrase("")
 
 
 class SearchFieldTransformer(TreeTransformer):
@@ -57,10 +111,24 @@ class SearchFieldTransformer(TreeTransformer):
         term_name = self._mapping.get(node.name, node.name)
         field_value_mapper = None
 
+        # TODO if you need another if clause here, please refactor
         if isinstance(term_name, FieldValueMapper):
             field_value_mapper = term_name
             term_name = field_value_mapper.term_name
-
+        if isinstance(term_name, RestrictedTermValue):
+            field_value_mapper = term_name
+            term_name = node.name
+        if isinstance(term_name, RestrictedTerm):
+            allows = term_name.allows(context["identity"])
+            term_name = node.name
+            # field_value_mapper is left as None on purpose - if the permission
+            # allows, we don't map any query, we allow it "as is"
+            if not allows:
+                raise QuerystringValidationError(
+                    _("Invalid search field: {field_name}.").format(
+                        field_name=node.name
+                    )
+                )
         # If a allow list exists, the term must be allowed to be queried.
         if self._allow_list and not term_name in self._allow_list:
             raise QuerystringValidationError(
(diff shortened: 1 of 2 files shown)


invenio_records_resources/services/records/queryparser/transformer.py
```diff
@@ -33,15 +33,55 @@ def term_name(self):
...
+    def __init__(self, permission, term_name=None, word=None, phrase=None):
+        self.permission = permission
+        self._word_fun = word
+        self._phrase_fun = phrase
+        self._term_name = term_name
+
+    def allows(self, identity):
+        if not (self._word_fun or self._phrase_fun):
+            return self.permission.allows(identity)
+        else:
```
kpsherva, line 57: this part a little bit flaky, I didn't have a better idea at the moment

invenio_records_resources/services/records/queryparser/transformer.py
```diff
@@ -56,11 +96,19 @@ def visit_search_field(self, node, context):
...
-
         if isinstance(term_name, FieldValueMapper):
             field_value_mapper = term_name
             term_name = field_value_mapper.term_name
-
+        if isinstance(term_name, RestrictedTerm):
+            field_value_mapper = term_name
+            allows = term_name.allows(context["identity"])
+            term_name = node.name
+            if not allows:
```
kpsherva, line 106: this is where we fail based on permissions if the query is not rewritten

invenio_records_resources/services/records/queryparser/transformer.py
```diff
@@ -33,15 +33,55 @@ def term_name(self):
         """Get the term name."""
         return self._term_name
 
-    def map_word(self, node):
+    def map_word(self, node, **kwargs):
```
kpsherva, line 36: for backwards compat

invenio_records_resources/services/records/queryparser/transformer.py
```diff
@@ -33,15 +33,57 @@ def term_name(self):
...
+
+    def map_word(self, node, context, **kwargs):
+        """Modify a word node."""
+        is_restricted = not self.permission.allows(context["identity"])
+        if self._word_fun and is_restricted:
+            return self._word_fun(node)
+        else:
+            return node
+
+    #
```
zzacharo, line 77: leftover?

invenio_records_resources/services/records/queryparser/transformer.py
zzacharo: That is not true no? Basically we change the signature and if someone has implemented their class and configured it on their instance then it will give an error...

invenio_records_resources/services/records/queryparser/transformer.py
```diff
@@ -79,9 +129,9 @@ def visit_search_field(self, node, context):
     def visit_word(self, node, context):
         """Visit a word term."""
         mapper = context.get("field_value_mapper")
-        yield node if mapper is None else mapper.map_word(node)
+        yield node if mapper is None else mapper.map_word(node, context=context)
```
zzacharo, line 132: that is a breaking change I think...see my comment above.

zzacharo (approved):

invenio_records_resources/services/records/queryparser/transformer.py
```diff
@@ -33,15 +33,57 @@ def term_name(self):
...
+
+    @property
+    def term_name(self):
+        """Get the term name."""
+        return self._term_name
+
+    def map_word(self, node, context, **kwargs):
+        """Modify a word node."""
+        is_restricted = not self.permission.allows(context["identity"])
+        if self._word_fun and is_restricted:
```
ntarocco, line 72: If it is restricted, and I don't have defined a `word_fun`, is that accepted?

invenio_records_resources/services/records/queryparser/transformer.py
```diff
@@ -33,15 +33,57 @@ def term_name(self):
...
 
 
+class RestrictedTerm:
+    """Class used to apply specific permissions to search."""
+
+    def __init__(self, permission, term_name=None, word=None, phrase=None):
+        """Constructor."""
+        self.permission = permission
+        self._word_fun = word
+        self._phrase_fun = phrase
```
ntarocco, line 52:
```suggestion
        self._word_func = word
        self._phrase_func = phrase
```
and I would change the above too

invenio_records_resources/services/records/queryparser/transformer.py
```diff
@@ -33,15 +33,57 @@ def term_name(self):
...
         return self._word_fun(node) if self._word_fun else node
 
-    def map_phrase(self, node):
+    def map_phrase(self, node, **kwargs):
         """Modify a phrase node."""
         return self._phrase_fun(node) if self._phrase_fun else node
 
 
+class RestrictedTerm:
+    """Class used to apply specific permissions to search."""
```
ntarocco, line 46: Missing a docstring: how should I define word/phrase doc strings?

invenio_records_resources/services/records/queryparser/transformer.py
```diff
@@ -56,11 +98,19 @@ def visit_search_field(self, node, context):
...
+        if isinstance(term_name, RestrictedTerm):
+            field_value_mapper = term_name
+            allows = term_name.allows(context["identity"])
+            term_name = node.name
+            if not allows:
+                raise QuerystringValidationError(
+                    _("Invalid search field: {field_name}.").format(
+                        field_name=node.name
+                    )
+                )
```
ntarocco, line 113: I would move the assignments afterwards, in the success branch, to minimise noise
```suggestion
            if not  term_name.allows(context["identity"]):
                raise QuerystringValidationError(
                    _("Invalid search field: {field_name}.").format(
                        field_name=node.name
                    )
                )
                
            field_value_mapper = term_name
            term_name = node.name
```

invenio_records_resources/services/records/queryparser/transformer.py
```diff
@@ -33,15 +33,57 @@ def term_name(self):
...
+            return self.permission.allows(identity)
+        else:
+            # if we specified the rewrite phrase of word function,
+            # let them take over the permission check
+            return True
+
+    @property
+    def term_name(self):
+        """Get the term name."""
+        return self._term_name
```
ntarocco, line 67: In the constructor, it is None by default.
Do we want to keep this?

invenio_records_resources/services/records/queryparser/transformer.py
ntarocco: Honestly, this assumes that you know a lot how to call the class, the name is also confusing to me. Plus missing a nice docstring :)

While the FieldValueMapper maps a search term to a specific field, as a shortcut (we should have better described it in the class), this one restricts searches.

What about splitting RestrictedTerm in 2, with specific behaviours instead?
```
"internal_notes.timestamp": RestrictedField(administration_permission),  # does not accept other params
"_exists_": RestrictedValue( administration_permission, word=word_internal_notes),   # requires word or phrase
```
because, in the end, we restrict a field, or a value (which happens to be a field when using the `_exist_` search operator).

This should allow you to have better class implementations, without mixing them with that `allows` and the weird if inside.

invenio_records_resources/services/records/queryparser/transformer.py
```diff
@@ -33,15 +33,57 @@ def term_name(self):
...
+
+    def map_word(self, node, context, **kwargs):
+        """Modify a word node."""
+        is_restricted = not self.permission.allows(context["identity"])
```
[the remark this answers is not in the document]
kpsherva: we cannot move it to the constructor, since context is not available at the init time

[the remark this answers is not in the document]
ntarocco: OK, but then, depending on the permissions evaluations, it could by cached by Identity to avoid re-evaluation.
If it is simply a `set()` `intersection`, then no need. However, is it will go through all the generators to have the final `set`, I would really evaluate caching it. To be checked?

invenio_records_resources/services/records/queryparser/transformer.py
```diff
@@ -60,7 +114,18 @@ def visit_search_field(self, node, context):
         if isinstance(term_name, FieldValueMapper):
             field_value_mapper = term_name
             term_name = field_value_mapper.term_name
-
+        if isinstance(term_name, RestrictedTermValue):
+            field_value_mapper = term_name
+            term_name = node.name
+        if isinstance(term_name, RestrictedTerm):
+            allows = term_name.allows(context["identity"])
+            term_name = node.name
```
ntarocco, line 122: Missing `field_value_mapper` assignment? move both after the if/raise?

ntarocco (approved):

PR #609, opened 2025-02-20
urls: introduce invenio_url_for-compatible links

fenekku:
Part of link generation feature - see https://github.com/inveniosoftware/invenio-base/pull/183

Depends on:
- https://github.com/inveniosoftware/invenio-base/pull/183 merged and released
- https://github.com/inveniosoftware/invenio-app/pull/98 merged and released

Not required for initial [invenio-rdm-records PR](https://github.com/inveniosoftware/invenio-rdm-records/pull/1960) - we can introduce the constructs from this PR into it later. But https://github.com/inveniosoftware/invenio-rdm-records/pull/1962 is required before this is released.

The high number of changed files is just because of the `tests/__init__.py` pattern (see https://github.com/inveniosoftware/invenio-rdm-records/pull/1954).

A guided commentary tour is provided.

### DIFF
--- a/invenio_records_resources/services/base/__init__.py
+++ b/invenio_records_resources/services/base/__init__.py
@@ -10,12 +10,21 @@
 """Base Service API."""
 
 from .config import ServiceConfig
-from .links import ConditionalLink, Link, LinksTemplate, NestedLinks
+from .links import (
+    ConditionalLink,
+    EndpointLink,
+    ExternalLink,
+    Link,
+    LinksTemplate,
+    NestedLinks,
+)
 from .results import ServiceItemResult, ServiceListResult
 from .service import Service
 
 __all__ = (
     "ConditionalLink",
+    "EndpointLink",
+    "ExternalLink",
     "Link",
     "LinksTemplate",
     "Service",
--- a/invenio_records_resources/services/files/config.py
+++ b/invenio_records_resources/services/files/config.py
@@ -10,13 +10,11 @@
 """Record Service API."""
 
 from ..base import ServiceConfig
-from ..records.links import RecordLink
 from .components import (
     FileContentComponent,
     FileMetadataComponent,
     FileProcessorComponent,
 )
-from .links import FileLink
 from .processors import ImageMetadataExtractor
 from .results import FileItem, FileList
 from .schema import FileSchema
@@ -41,14 +39,9 @@ class FileServiceConfig(ServiceConfig):
 
     max_files_count = 100
 
-    file_links_list = {
-        "self": RecordLink("{+api}/records/{id}/files"),
-    }
-
-    file_links_item = {
-        "self": FileLink("{+api}/records/{id}/files/{+key}"),
-        "content": FileLink("{+api}/records/{id}/files/{+key}/content"),
-    }
+    # Inheriting resource config should define these
+    file_links_list = {}
+    file_links_item = {}
 
     components = [
         FileMetadataComponent,
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -16,9 +16,9 @@ fixtures are available.
 import pytest
 from flask_principal import Identity, Need, UserNeed
 from invenio_app.factory import create_api as _create_api
-from mock_module.config import MockFileServiceConfig, ServiceConfig
 
 from invenio_records_resources.services import FileService, RecordService
+from tests.mock_module.config import FileServiceConfig, ServiceConfig
 
 pytest_plugins = ("celery.contrib.pytest",)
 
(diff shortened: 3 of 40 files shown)


invenio_records_resources/services/files/config.py
```diff
@@ -41,14 +39,10 @@ class FileServiceConfig(ServiceConfig):
...
-        "self": RecordLink("{+api}/records/{id}/files"),
-    }
-
-    file_links_item = {
-        "self": FileLink("{+api}/records/{id}/files/{+key}"),
-        "content": FileLink("{+api}/records/{id}/files/{+key}/content"),
-    }
+    # Inheriting resource config should define these
+    # TODO: Edit rdm-records such that it doesn't rely on this first
+    file_links_list = {}
```
fenekku, line 43: See https://github.com/inveniosoftware/invenio-rdm-records/pull/1962

tests/conftest.py
```diff
@@ -56,13 +58,18 @@ def extra_entry_points():
...
-            "mock_module = mock_module.jsonschemas",
+            "mock_module = tests.mock_module.jsonschemas",
         ],
         "invenio_search.mappings": [
-            "records = mock_module.mappings",
+            "records = tests.mock_module.mappings",
+        ],
+        "invenio_base.api_blueprints": [
+            "mock_module_mocks = tests.mock_module:create_mocks_bp",
+            # still present even though above doesn't support files
```
fenekku, line 71: which is per the original tests

tests/resources/conftest.py
```diff
@@ -33,11 +33,11 @@ def record_resource(service):
     return RecordResource(CustomRecordResourceConfig, service)
 
 
-@pytest.fixture(scope="module")
-def base_app(base_app, record_resource):
-    """Application factory fixture."""
-    base_app.register_blueprint(record_resource.as_blueprint())
-    yield base_app
```
fenekku, line 40: By relying on entry points to load blueprints, this kind of thing is not needed anymore.

tests/services/test_service.py
```diff
@@ -100,7 +98,11 @@ def test_read_many_pid_values(app, search_clear, service, identity_simple, input
     Record.index.refresh()
 
     records = service.read_many(
-        identity_simple, ids=[item_one.id, item_two.id], fields=["metadata.type.type"]
+        identity_simple,
+        ids=[item_one.id, item_two.id],
+        # NOTE: "id" and "pid" are needed to produce the links in the result
+        #       Previous link generation would silently produce incorrect links
+        fields=["id", "pid", "metadata.type.type"],
```
fenekku, line 105: Discussion: yeah this is good and bad. Essentially the new way of generating links informs us of cases where urls were malformed... by throwing a BuildError. I don't think "id" and "pid" are universal though... If they are, we could add them to https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/records/service.py#L440 or make those default dumper fields configurable in `RecordServiceConfig`.

invenio_records_resources/services/base/__init__.py
```diff
@@ -10,13 +10,14 @@
 """Base Service API."""
 
 from .config import ServiceConfig
-from .links import ConditionalLink, Link, LinksTemplate, NestedLinks
+from .links import ConditionalLink, Link, Link2, LinksTemplate, NestedLinks
```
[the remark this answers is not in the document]
max-moser: just to start everybody's favorite topic: naming

from the comments, it sounds a bit like the idea is to prepare the new infrastrucutre as backwards-compatible as possible, and then make a switch in a series of major releases (i.e. rename the new infra to the old names and kick the old ones out)?
in that case, the new names would be temporary anyway and it wouldn't matter too much what they're called

otherwise, maybe we could encode more of the intention behind the classes in their naming
e.g. `Link` -> `ExternalLink` (with an alias `Link` pointing to it for backwards-compat), `Link2` -> `InvenioLink`

what do you think?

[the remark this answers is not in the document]
fenekku: I like this:

> e.g. Link -> ExternalLink (with an alias Link pointing to it for backwards-compat), Link2 -> InvenioLink

That would save the dance of switching to a new name and then back to the other, but-not-really-because-still-useful... `ExternalLink` is indeed the right name for what that one should be used for moving forward. `InvenioLink` is fine by me (I never know if we should add the "RDM" part, but without is good by me).

I will address that tomorrow :+1:

max-moser (approved):
i just took it for a test ride in a local dev v13 (only installed this PR, none of the other remaining ones).
creation of records (with & without files) and communities, and community-submission requests work as expected

PR #610, opened 2025-02-25
files: remove HTTP code from read_content

egabancho:
* Setting the return code directly hides the return code that ``send_file`` might be issuing, resulting in odd situations when the storage layer returns a 302, i.e. offload download.

---

When an instance uses local storage, eventually [send_file](https://github.com/inveniosoftware/invenio-files-rest/blob/master/invenio_files_rest/helpers.py#L62) gets called and the file is streamed through the application server.

If one uses Invenio-S3,  instead of serving the file directly, the application [issues a redirect](https://github.com/inveniosoftware/invenio-s3/blob/master/invenio_s3/helpers.py#L20) to a short-lived signed URL, relaying the work of streaming the file to the storage server, offloading.

The problem is that send_file itself returns a response with its own response code, etc. The 200 [here](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/resources/files/resource.py#L204) basically overwrites that response code and the redirect doesn't work.

We don't experience that same behavior with EOS because the redirect is happening at [nginx level](https://github.com/zenodo/zenodo-rdm/blob/master/site/zenodo_rdm/files.py#L74) using X-Accel-Redirect.

### DIFF
--- a/invenio_records_resources/resources/files/resource.py
+++ b/invenio_records_resources/resources/files/resource.py
@@ -201,7 +201,7 @@ class FileResource(ErrorHandlersMixin, Resource):
         if obj is not None and emitter is not None:
             emitter(current_app, record=item._record, obj=obj, via_api=True)
 
-        return item.send_file(), 200
+        return item.send_file()
 
     @request_view_args
     def read_archive(self):


egabancho:
This should probably fix https://github.com/inveniosoftware/invenio-app-rdm/issues/2017 🤔

tmorrell:
This broke uploads on my (fairly ancient) test system 

```
flask_resources/responses.py", line 40, in inner
   return resource_requestctx.response_handler.make_response(*res, many=many)
TypeError: make_response() got multiple values for argument 'many'

```

and didn't seem to fix the content issue. I'll try to test on v13 soon and see if it works better there.

egabancho:
I tested it locally on `v13.0.0b1.dev25`, and I didn't get any error. 
How did you test it? Did you check-pick the change?

slint (approved):

tmorrell (approved):
I've now tested it on v13.0.0b2 and didn't run into any issues. It fixed the content issues on AWS S3 and is good to go!

PR #615, opened 2025-04-10
service: records: add MultiFieldsResolver

yashlamba:
Moved from https://github.com/inveniosoftware/invenio-requests/pull/431
Closes https://github.com/inveniosoftware/invenio-requests/issues/428

### DIFF
--- a/invenio_records_resources/services/records/results.py
+++ b/invenio_records_resources/services/records/results.py
@@ -448,11 +448,7 @@ class FieldsResolver:
         The `id` field used to match the resolved record is hardcoded,
         as in the `read_many` method.
         """
-        fields = []
-        for field in self._fields:
-            if field.has(service, value):
-                fields.append(field)
-        return fields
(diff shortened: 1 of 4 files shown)


invenio_records_resources/services/records/results.py
```diff
@@ -511,3 +506,77 @@ def expand(self, identity, hit):
...
+    """Resolve the reference record for each of the configured fields.
+
+    Given a list of fields referencing other records/objects,
+    it fetches and returns the dereferenced record/obj.
+
+    This class supports resolution of nested fields and efficiently batches
+    resolution requests to services.
+    """
+
+    def _collect_values(self, hits):
```
zzacharo, line 521: nit: If there are any common code blocks here I would extract them in the base class so we can reuse them here and keep only the needed modified parts to support the list of things

zzacharo (approved):

PR #638, opened 2025-06-26
fix: pkg_resources DeprecationWarning

utnapischtim:
* UserWarning: pkg_resources is deprecated as an API. See
  https://setuptools.pypa.io/en/latest/pkg_resources.html. The
  pkg_resources package is slated for removal as early as 2025-11-30.
  Refrain from using this package or pin to Setuptools<81.

### DIFF
--- a/invenio_records_resources/services/files/processors/image.py
+++ b/invenio_records_resources/services/files/processors/image.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2021-2024 CERN.
+# Copyright (C) 2025 Graz University of Technology.
 #
 # Invenio-Records-Resources is free software; you can redistribute it and/or
 # modify it under the terms of the MIT License; see LICENSE file for more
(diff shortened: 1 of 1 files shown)


invenio_records_resources/services/files/processors/image.py
```diff
@@ -1,24 +1,24 @@
...
 
 try:
-    pkg_resources.get_distribution("wand")
     from wand.image import Image
 
     HAS_IMAGEMAGICK = True
-except pkg_resources.DistributionNotFound:
+except ImportError:
     # Python module not installed
     HAS_IMAGEMAGICK = False
```
fenekku, line 23: AFter looking this up a bit, I can't see how that former code would have checked if the underlying ImageMagick library was there; it still only tested for the Wand distribution (i.e. the Python bindings). So removal and checking the import make sense to me — it's the best we can do. (and we do check for requirments via invenio-cli anyway).

Let's remove the duplicate `except ImportError` below though.

PR #647, opened 2025-07-26
refactor: remove context usage

utnapischtim:
* marshmallow deprecated the context usage.

* this commit introduces two ContextVars to replace the marshmallow context

### DIFF
--- a/invenio_records_resources/services/records/schema.py
+++ b/invenio_records_resources/services/records/schema.py
@@ -2,17 +2,21 @@
 #
 # Copyright (C) 2020-2024 CERN.
 # Copyright (C) 2022 TU Wien.
+# Copyright (C) 2025 Graz University of Technology.
 #
 # Invenio-Records-Resources is free software; you can redistribute it and/or
 # modify it under the terms of the MIT License; see LICENSE file for more
 # details.
 
 """Record schema."""
+
 from copy import deepcopy
 from datetime import timezone
 
+from invenio_access.context import context_identity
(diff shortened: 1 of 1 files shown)


invenio_records_resources/services/records/schema.py
```diff
@@ -64,30 +68,23 @@ def __init__(self, service, schema):
         # TODO: Change constructor to accept a permission_policy_cls directly
         self._permission_policy_cls = service.config.permission_policy_cls
 
-    def _build_context(self, base_context):
-        context = {**base_context}
-        default_identity = context["identity"]  # identity required in context
+    def set_field_permission_check(self, context):
+        default_identity = context_identity.get()
```
utnapischtim, line 72: this one is wrong, it does not use the `context["identity"]` as the first option, `context_identity.get()` should be the fallback

invenio_records_resources/services/records/schema.py
```diff
@@ -100,6 +97,5 @@ def load(self, data, schema_args=None, context=None, raise_errors=True):
     def dump(self, data, schema_args=None, context=None):
         """Dump data using wrapped schema and dynamic schema_args + context."""
         schema_args = schema_args or {}
-        base_context = context or {}
-        context = self._build_context(base_context)
-        return self.schema(context=context, **schema_args).dump(data)
+        self.set_field_permission_check(context)
```
utnapischtim, line 100: wrap it with a ContextManager

utnapischtim:
closed in favor of #653 

if the problems should be handled independently, please update this PR

PR #652, opened 2025-09-25
fix(service): track errors on commit for create_update_many

jrcastro2:


invenio_records_resources/services/records/service.py
```diff
@@ -651,13 +651,17 @@ def _create_record(record_dict):
...
-            for _, record, errors, exc in records_processed
-            if errors == [] and exc is None
-        ]
-        uow.register(RecordBulkCommitOp(records, self.indexer))
+        valid_records = []
+
+        for i, (action, record, errors, exc_) in enumerate(records_processed):
+            if errors == [] and exc_ is None:
+                try:
+                    record.commit()
```
ntarocco, line 659: We normally use the `uow` to avoid committing immediately, but instead committing at the end of the service method.

invenio_records_resources/services/records/service.py
```diff
@@ -651,13 +651,17 @@ def _create_record(record_dict):
...
+                    valid_records.append(record)
+                except Exception as exc:
+                    # If commit fails we register the error for this record
+                    records_processed[i] = (action, record, errors, exc)
```
ntarocco, line 663: is there any value in doing this? What are we gaining?

invenio_records_resources/services/records/service.py
ntarocco, line 659: can you please add a comment that explains that we cannot use the `uow` because of the error tracking?

invenio_records_resources/services/records/service.py
```diff
@@ -651,13 +651,17 @@ def _create_record(record_dict):
...
-        ]
-        uow.register(RecordBulkCommitOp(records, self.indexer))
+        valid_records = []
+
+        for i, (action, record, errors, exc_) in enumerate(records_processed):
+            if errors == [] and exc_ is None:
+                try:
+                    record.commit()
+                    valid_records.append(record)
+                except Exception as exc:
```
ntarocco, line 661: Can we have a more specific exception here instead?

invenio_records_resources/services/records/service.py
jrcastro2: Yes, added, thanks!

invenio_records_resources/services/records/service.py
jrcastro2: I added the base `RelationError`, to my understanding all validation errors happening on commit for the system field should inherit from this. This works at the moment, let me know if I also should add the general Exception or just keep it like this.

PR #677, opened 2026-04-10
feat(facets): configurable `post_filter` and dynamic bounds for `DateFacet`

slint:
> AI use disclosure: I developed majority of the feature with Claude Code Opus 4.6. Most of corrections/guidance were related to naming, backwards compat, concerns on thread safety (since we're "storing" the selected facet values on the facet object, which turned out to be fine since we're deepcopy-ing things on every request). 

WIP: I would like to put a before/after video in the companion PR at https://github.com/inveniosoftware/react-searchkit/pull/297 to showcase the new behavior.

- **feat(facets): add Facet base class with post_filter and prepare_aggregation**
  * Add Facet base class (subclassing dsl.Facet) with:
    - ``post_filter`` attribute controlling filter application strategy
    - ``prepare_aggregation()`` hook called before ``get_aggregation()``
  * Update TermsFacet and DateFacet to inherit from Facet
  * Split FacetsParam.filter() into post_filters and direct_filters
  * Add FacetsParam._combine() helper
  * Call prepare_aggregation() during aggregation phase
  

- **feat(facets): add hard_bounds and dynamic bounds to DateFacet**
  * Add optional ``hard_bounds`` param to DateFacet for date_histogram
    aggregations (e.g. ``hard_bounds={"min": "1800", "max": "now/y"}``)
  * Implement ``prepare_aggregation()`` to store active filter values
  * Add ``_effective_bounds()`` to dynamically replace hard_bounds with
    the filter range when it extends outside configured bounds
  * Bounds computation is side-effect free (no mutation of _hard_bounds)

### DIFF
--- a/invenio_records_resources/services/records/facets/facets.py
+++ b/invenio_records_resources/services/records/facets/facets.py
@@ -16,6 +16,30 @@ from dateutil.parser import isoparse
 from invenio_search.engine import dsl
 
 
+class Facet(dsl.Facet):
+    """Base facet with filtering behavior defaults.
+
+    Subclasses can override ``post_filter`` to control whether their filter is
+    applied as a post_filter (preserves aggregation counts from the full result
+    set) or as a direct query filter (aggregations reflect filtered results).
+
+    ``post_filter`` can be set as a class attribute, passed to the constructor,
+    or set on an instance after construction.
+    """
+
+    post_filter = True
+
+    def __init__(self, post_filter=None, **kwargs):
+        """Constructor."""
+        if post_filter is not None:
+            self.post_filter = post_filter
+        super().__init__(**kwargs)
+
+    def prepare_aggregation(self, filter_values):
+        """Prepare the aggregation based on active filter values. No-op by default."""
+        pass
+
+
 class LabelledFacetMixin:
     """Mixin class for overwriting the default get_values() method."""
 
@@ -73,7 +97,7 @@ class LabelledFacetMixin:
         return {"buckets": out, "label": str(self._label)}
 
 
-class TermsFacet(LabelledFacetMixin, dsl.TermsFacet):
+class TermsFacet(LabelledFacetMixin, Facet, dsl.TermsFacet):
     """Terms facet.
 
     .. code-block:: python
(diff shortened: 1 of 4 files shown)


invenio_records_resources/services/records/params/facets.py
```diff
@@ -31,31 +31,45 @@ def facets(self):
...
-            return
-        # Store filter.
-        self._filters[name] = f
+        if f is not None:
+            self._filters[name] = f
+
+    @staticmethod
+    def _combine(filters):
+        """Combine filters with AND."""
+        combined = filters[0]
```
OliverGeneser, line 44: We should check that filters is not empty

invenio_records_resources/services/records/facets/facets.py
```diff
@@ -542,15 +573,94 @@ def __init__(
...
+        length = self._FORMAT_SLICE.get(self._format, 4)
+        return normalized_date[:length]
+
+    def _resolve_bound(self, value, is_start):
+        """Resolve a bounds value to a comparable YYYY-MM-DD string.
+
+        Returns None for dynamic expressions like ``"now/y"`` since they
+        should never be considered exceeded by a filter.
+        """
+        if not value or value.startswith("now"):
```
OliverGeneser: What if the value is not a string? We should add a isinstance check

invenio_records_resources/services/records/facets/facets.py
```diff
@@ -542,15 +573,94 @@ def __init__(
...
         super().__init__(label=label, **kwargs)
 
+    def prepare_aggregation(self, filter_values):
+        """Prepare the aggregation by storing active filter values.
+
+        Called before ``get_aggregation`` so that ``_effective_bounds`` can
+        adjust ``hard_bounds`` when the filter range extends outside the
+        configured bounds.
+        """
+        self._active_filter_values = filter_values
```
OliverGeneser, line 583: Hmm this makes the facet stateful... Should we instead pass the filter values explicitly to get_aggregation() rather than storing them in the instance?

invenio_records_resources/services/records/facets/facets.py
slint: This is called with values from `._hard_bounds`... which I guess probably accepts `"min": 1800`... I'll add the check and cast here.

invenio_records_resources/services/records/facets/facets.py
slint: This was a bit of a struggle, because the base OpenSearch `Facet` class, doesn't accept any params, and I don't want to do a `inspect.signature` to check if we use our "own" sub-class/version of it.

We deepcopy the facets on each request, so in practice we have a "fresh" copy to modify everytime 🙃

PR #678, opened 2026-04-13
feat(facets): add `bounds_mode` to DateFacet and fix year padding

slint:
* Fixed a bug where filter values like ``0190..0500`` were passed as ``190-01-01`` / ``500-12-31`` to the query, causing OpenSearch to error with "failed to parse date field" since strict ISO requires a 4-digit year.
* Add ``match_filter_bounds`` flag to control how ``hard_bounds`` interacts with the active filter.
* Default ``False`` keeps the configured bounds when the filter is inside; replaces with the filter range when outside.
* Set to ``True`` to always use the filter range when a filter is active, focusing the histogram on the user's selection.
---

### `match_filter_bounds=True` (i.e. the default behavior)

CleanShot 2026-04-29 at 09 02 33

### `match_filter_bounds=False`

CleanShot 2026-04-29 at 09 05 59

### DIFF
--- a/invenio_records_resources/services/records/facets/facets.py
+++ b/invenio_records_resources/services/records/facets/facets.py
@@ -10,7 +10,7 @@
 """Facets types defined."""
 
 import re
-from datetime import date
+from datetime import date, timedelta
 
 from dateutil.parser import isoparse
 from invenio_search.engine import dsl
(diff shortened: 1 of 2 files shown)


invenio_records_resources/services/records/facets/facets.py
```diff
@@ -598,49 +600,56 @@ def get_aggregation(self):
...
+        """Check if a normalized filter range extends outside hard_bounds."""
+        if not self._hard_bounds:
+            return False
+        bounds_min = self._resolve_bound(self._hard_bounds.get("min"), True)
+        bounds_max = self._resolve_bound(self._hard_bounds.get("max"), False)
+        start, end = normalized_filter["start"], normalized_filter["end"]
+        return (start and bounds_min and start < bounds_min) or (
+            end and bounds_max and end > bounds_max
+        )
+
```
zubeydecivelek, line 668: We only check start < bounds_min / end > bounds_max and ignore start_inclusive / end_inclusive. Do you think it makes sense to use the start_inclusive / end_inclusive?

zubeydecivelek (approved):
Looks good to me :)

invenio_records_resources/services/records/facets/facets.py
slint: Wise words! It's a bit trickier than I thought though, because e.g. if you ask for `(1970..]` you basically mean `[1971..]` and need to do this conversion for a few cases (and hopefully it's understandable when one reads the code 😅). I'll have a go at it and ping back here :)

PR #680, opened 2026-04-16
service: add support for search parameters (e.g. size, scroll, etc.)

ptamarit:
Add support for the `scroll` duration parameter.
By default, [`opensearch-py` sets a value a `5m` (5 minutes)](https://github.com/opensearch-project/opensearch-py/blob/v2.8.0/opensearchpy/_async/helpers/actions.py#L318), but setting a higher value might be useful for slow queries.

### DIFF
--- a/invenio_records_resources/services/records/service.py
+++ b/invenio_records_resources/services/records/service.py
@@ -128,6 +128,7 @@ class RecordService(Service, RecordIndexerMixin):
         preference=None,
         extra_filter=None,
         versioning=True,
+        size=None,
         scroll=None,
     ):
         """Instantiate a search class."""
@@ -161,6 +162,9 @@ class RecordService(Service, RecordIndexerMixin):
                 .params(version=True)
             )
 
+        if size:
+            search = search.params(size=size)
+
         if scroll:
             search = search.params(scroll=scroll)
 
@@ -181,6 +185,7 @@ class RecordService(Service, RecordIndexerMixin):
         extra_filter=None,
         permission_action="read",
         versioning=True,
+        size=None,
         scroll=None,
     ):
         """Factory for creating a Search DSL instance."""
@@ -192,6 +197,7 @@ class RecordService(Service, RecordIndexerMixin):
             preference=preference,
             extra_filter=extra_filter,
             versioning=versioning,
+            size=size,
             scroll=scroll,
         )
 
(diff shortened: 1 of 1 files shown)


invenio_records_resources/services/records/service.py
```diff
@@ -128,6 +128,8 @@ def create_search(
         preference=None,
         extra_filter=None,
         versioning=True,
+        size=None,
+        scroll=None,
```
slint, line 132: minor: should we maybe just have an `extra_params=None`, that we pass to `search.params(**extra_params)`?

ptamarit: I named it `search_parameters` (instead of `extra_parameters`) to follow the existing `search_preference` argument.

In the end, I only need `scroll` and not `size`, but this is a more flexible solution for potential future use cases and avoids bloating with more and more arguments.

invenio_records_resources/services/records/service.py
```diff
@@ -201,6 +208,7 @@ def _search(
         identity,
         params,
         search_preference,
+        search_params=None,
```
ptamarit, line 211: Here, I did not fully follow the `search_preference` pattern, and made `search_params`.
Even though `_search` is an internal method, there is a risk that making this parameter mandatory breaks some code.

invenio_records_resources/services/records/service.py
slint: TBH, I dislike the `search_preference` arg, because it's such a specific/obscure behavior that we want to control, but somehow it has managed to propagate throughout all our signatures, instead of being automatically handled/injected with a sane default.

Given that we have normal `params` (which are coincidentally named that way and have nothing to do with what we pass to `Search.params(...)`, see below), I think `extra_params` matches the `extra_filter` as "last resort" kind of way to hook-into building the search query.

---

Looking a bit closer at the code now, I think we could have also implemented this as a `params_interpreters_cls` in the search options. At the same time, this exposes it to end-users (i.e. anyone can pass `/api/records?scroll=...`) and I don't see an explicit way to differentiate between internal/public parameters.

ptamarit:
Alternative PR with changes only in Zenodo (where it's currently needed):
zenodo/zenodo-rdm#1367

ptamarit:
Closing this PR for now since the change was in Zenodo where it was needed zenodo/zenodo-rdm#1367.

PR #682, opened 2026-05-07
files: limit archive download by total file size

slint:


invenio_records_resources/resources/errors.py
```diff
@@ -192,6 +193,12 @@ class ErrorHandlersMixin:
                 description=str(e),
             )
         ),
+        ArchiveDownloadTooLargeException: create_error_handler(
+            lambda e: HTTPJSONException(
+                code=400,
```
OliverGeneser, line 198: Shouldn't we use 413 Content Too Large?

Image

invenio_records_resources/services/files/service.py
```diff
@@ -113,6 +117,25 @@ def list_files(self, identity, id_):
...
+
+        Same as ``list_files`` but rejects records whose total file size
+        exceeds ``RECORDS_RESOURCES_ARCHIVE_DOWNLOAD_MAX_SIZE``.
+        """
+        record = self._get_record(id_, identity, "read_files")
+        max_size = current_app.config.get("RECORDS_RESOURCES_ARCHIVE_DOWNLOAD_MAX_SIZE")
+        if (
+            max_size is not None
+            and record.files.enabled
+            and record.files.total_bytes > max_size
```
OliverGeneser, line 156: Is total_bytes something we calculate on upload, or is it something we aggregated on a schedule? If it's the latter then this might throw a TypeError

OliverGeneser (approved):

invenio_records_resources/resources/errors.py
slint: I thought about it, but after reading also [the MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/413), 413 is more common for clients sending data, and not on the receiving side.

invenio_records_resources/services/files/service.py
slint: It's a computed field... Funnily, it does the `sum` every time we call it 😅 Though I think it's not worth caching.

PR #683, opened 2026-05-16
Staged file uploads, releasing DB connection

slint:
- **feat(files): add staged-local transfer for split-phase uploads**
  Splits the byte transfer out of the DB transaction so the
  connection pool is no longer the upload-concurrency cap. Opt in
  via ``RECORDS_RESOURCES_USE_STAGED_TRANSFER`` (default ``False``).
  

- **fix(files): hide pending staged uploads from archive, dumper, files manager**
  ``File.from_dump`` now sets ``readable=True`` so loaded-from-search
  records aren't treated as pending under the new predicate.
  

- **tests(files): check staged uploads release the DB connection during the upload**
  

- **fix(files): clean up rows when a staged upload fails mid-stream**
  

- **feat(files): add staged-fetch transfer**
  

- **tests(files): in-flight config-rollback for staged uploads**


invenio_records_resources/config.py
```diff
@@ -58,3 +58,7 @@
 
 RECORDS_RESOURCES_ARCHIVE_DOWNLOAD_MAX_SIZE = None
 """Max total file size (bytes) for archive download. ``None`` disables the cap."""
+
+
+RECORDS_RESOURCES_USE_STAGED_TRANSFER = False
```
OliverGeneser: This is intended to be a short-lived feature flag right? To me this is a clear resource lifetime fix, so I just want to make sure the plan is to eventually make this the default rather than keeping it behind a feature flag long-term 😅

slint: Yes, I would like this to be permanently enabled (and thus no feature flag necessary), but it's a larger shift, so I think taking it for a run on Zenodo production for at least a month or so will make us feel safer about changing for everyone.

invenio_records_resources/services/files/components/metadata.py
```diff
@@ -31,16 +34,32 @@ def init_files(self, identity, id_, record, data):
...
                 file_service=self.service,
                 key=file_metadata["key"],
-                transfer_type=file_metadata["transfer"]["type"],
+                transfer_type=transfer_data["type"],
                 uow=self.uow,
             )
+            kwargs = {}
+            if use_staged and transfer_data["type"] in (
+                LOCAL_TRANSFER_TYPE,
+                FETCH_TRANSFER_TYPE,
```
slint: I'm not extremely happy with these hardcoded transfer types here... In a previous iteration, this was delegated to the transfer type/providers implementations, but it didn't "feel" right in terms of polymorphism/SoC.

Maybe a better alternative would be to have them as config or as something attribute on the transfer classes.

invenio_records_resources/services/files/upload.py
slint: This chonker here handles all the different upload procedures. It's still very much service-layer stuff (most parts were moved from service methods), since it manages the unit of work as well.

invenio_records_resources/services/files/upload.py
```diff
@@ -0,0 +1,581 @@
...
+            )
+        except FileSizeError as error:
+            self._abort(record, file_key, error, cleanup_storage=False)
+            raise
+        except (ClientDisconnected, OSError):
+            error = TransferException(f'Transfer of File with key "{file_key}" failed.')
+            return (
+                self._abort(record, file_key, error, cleanup_storage=False),
+                error,
+            )
```
OliverGeneser: Hmm I'm a bit concerned that this is not all the potential exceptions where we should release the claim 🤔

invenio_records_resources/services/files/upload.py
```diff
@@ -0,0 +1,581 @@
...
+                return self._finalize(
+                    prepared,
+                    uri,
+                    size,
+                    checksum,
+                    retry=False,
+                    components=components,
+                    component_args=component_args,
+                )
+            raise
```
OliverGeneser: I don't like that there's only a single retry..

invenio_records_resources/services/files/upload.py
```diff
@@ -0,0 +1,581 @@
...
+                )
+
+                if (
+                    file_model.object_version_id != obj.version_id
+                    or obj.file_id != file_instance.id
+                ):
+                    raise TransferException(
+                        f'Upload for file "{prepared.file_key}" no longer owns '
+                        "the file."
+                    )
```
OliverGeneser: Hmm if the file_key no longer owns a file, shouldnlt _abort() then verify that the file record still points to the prepared upload before just deleting a file by key?

invenio_records_resources/services/files/service.py
```diff
@@ -471,14 +444,14 @@ def set_multipart_file_content(
             current_app.logger.exception("File upload transfer failed.")
             # we gracefully fail so that uow can commit the cleanup operation in
             # FileContentComponent
-            errors = "File upload transfer failed."
+            errors.append(_("File upload transfer failed."))
```
OliverGeneser: Is the change to a list of errors instead of a string intentional? Because I think it's the correct thing, but I'm not sure if some other module expects it to return a string 😅

OliverGeneser (changes requested):
Uploads and deletes works for me and the connection is correctly closed. 

I think Clanker might be mixing stuff, because from what I see we use `readable=False` to show that a file is not readable from local storage if it's a remote transfer, but now it's being used as pending/unavailable status 🤔

invenio_records_resources/services/files/upload.py
slint: They were already handled in the original service methods. Agree, we should check on Zenodoproduction Sentry to see which others we've come across and add here.

invenio_records_resources/services/files/upload.py
slint: In case this fails we either:

- or we send a Celery task anyway here that will do the clean-up in an idempotent way
- need a periodic/schedule Celery task to find dangling File Instances and clean them up

OliverGeneser (approved):
lgtm

OliverGeneser (changes requested):
Found an issue (Currently trying to reproduce)

OliverGeneser (approved):

fenekku:
@mesemus has looked at file uploads and database connections recently so better placed to take a look at this fully. (I won't have time to fully immerse into this one in near-term).

invenio_records_resources/services/files/components/metadata.py
ntarocco: One idea could be to move this part inside the new FileUpload transfer class, which can internally encapsulate the logic to check if `RECORDS_RESOURCES_USE_STAGED_TRANSFER` is enabled or not. It will also encapsulate there the check of transfer type which is anyway duplicated there.

Here we would see only:
```
FileUpload(...).init(...)
```
which might do the below or skip.

invenio_records_resources/services/files/upload.py
```diff
@@ -0,0 +1,687 @@
...
+
+class UploadConflict(TransferException):
+    """Another file operation got in the way of this upload."""
+
+
+class UploadSuperseded(UploadConflict):
+    """The rows this upload prepared are no longer the current ones."""
+
+
+class FileUpload:
```
ntarocco: Why not creating this object ones, setting it a the level of the service (or another existing object) and then call the instance `.fileupload`, instead of repeating `FileUpload(self)` around?

invenio_records_resources/services/files/upload.py
```diff
@@ -0,0 +1,687 @@
...
+    def __init__(self, service):
+        """Initialize the upload."""
+        self.service = service
+
+    def set_content(
+        self, identity, id_, record, file_key, stream, content_length, uow=None
+    ):
+        """Set file content using the upload path chosen during initialization."""
+        if uow is not None:
+            return self._set_content_in_uow(
```
ntarocco: Question: is this `if` added just in case something breaks?
It looks to me that we now want to always avoid passing the `uow`. Or maybe I missed where it is still used.

invenio_records_resources/services/files/upload.py
```diff
@@ -0,0 +1,687 @@
...
+        with UnitOfWork(db.session) as managed_uow:
+            remaining = self._delete_all_files_in_uow(
+                identity, id_, record, managed_uow
+            )
+            managed_uow.commit()
+        deleted.update({file.key: file for file in remaining})
+        return [deleted[file_key] for file_key in file_keys]
+
+    @staticmethod
+    def uses_staged_upload(file_record):
```
ntarocco: should this method also check for the config flag (and drop the static in this case)?

ntarocco (commented):
Nice work, I have reviewed only the high-level logic change. My feeling is that the new implementation might be overly complicated, and it could be simplified, in particular the FileUpload class.
However, LGTM.

invenio_records_resources/services/files/upload.py
```diff
@@ -0,0 +1,687 @@
...
+                record,
+                stream,
+                content_length,
+                uow=uow,
+            )
+
+        if self.uses_staged_upload(record.files.get(file_key)):
+            component_args = (identity, id_, file_key, stream, content_length)
+            before_components, after_components = self._split_content_components()
+            return self._upload(
```
ntarocco: Minor: I suggest a small refactoring to avoid the 3 return statements and make this different cases more explicit. It took me a while to see that the `_set_content_in_uow` was bot called if `staged`.

One way could be to have 2 different private methods: one that will do the staged upload, and the other that will do the classic upload.

mesemus:
I've had the first pass over the PR. The one concern I have is that it modifies the exact time when file instance is created and its content - which is then different to what goes on in other transfers and should be unified.

How fast do we need to have this merged? Would it be possible to go over the architecture and set some definitions that are blurred at the moment (the whole "when a file instance is created and what its existence means, what does the readable/writable exactly mean and how/when it should be set up, ...). We have an outstanding issue with remote files at the moment and my worries are that without the exact definitions we might close some doors/introduce more inconsistencies when we'll be fixing the remote files.

OliverGeneser:
> I've had the first pass over the PR. The one concern I have is that it modifies the exact time when file instance is created and its content - which is then different to what goes on in other transfers and should be unified.
> 
> How fast do we need to have this merged? Would it be possible to go over the architecture and set some definitions that are blurred at the moment (the whole "when a file instance is created and what its existence means, what does the readable/writable exactly mean and how/when it should be set up, ...). We have an outstanding issue with remote files at the moment and my worries are that without the exact definitions we might close some doors/introduce more inconsistencies when we'll be fixing the remote files.

In Zenodo we need it ASAP, currently deployed to Dev/Sandbox, as the file uploads is currently eating all of our PgBouncer connections :sweat_smile: 
Alex is currently on vacation, but he will be back 28th Sep. so we can deploy from this branch until he's back and then I think it makes sense to have a meeting about the definitions :+1:

ntarocco:
> > I've had the first pass over the PR. The one concern I have is that it modifies the exact time when file instance is created and its content - which is then different to what goes on in other transfers and should be unified.
> > How fast do we need to have this merged? Would it be possible to go over the architecture and set some definitions that are blurred at the moment (the whole "when a file instance is created and what its existence means, what does the readable/writable exactly mean and how/when it should be set up, ...). We have an outstanding issue with remote files at the moment and my worries are that without the exact definitions we might close some doors/introduce more inconsistencies when we'll be fixing the remote files.
> 
> In Zenodo we need it ASAP, currently deployed to Dev/Sandbox, as the file uploads is currently eating all of our PgBouncer connections 😅 Alex is currently on vacation, but he will be back 28th Sep. so we can deploy from this branch until he's back and then I think it makes sense to have a meeting about the definitions 👍

Agree, we can review/improve later on, no problem with that!

ptamarit:
Just rebased to fix conflicts:
* Removed this from `CHANGES.rst`:
    ```
    Version v11.1.0 (unreleased)
    
    - files: release database connections while local and fetch uploads stream to storage.
    ```
* Fixed a conflict in `invenio_records_resources/services/files/service.py` due to #724 which also introduced a constructor in `FileService`

ptamarit:
Some of the comments were addressed.
We deployed this to Zenodo QA for a few days and it is behind a feature flag, so we are proposing to merge/release and iterate from there.

PR #687, opened 2026-06-05
Refactor(schema): Update record schemas for singleton reuse

### LINKED ISSUE #3480: Marshmallow: Optimize schema instantiation to improve serialization performance
Currently Invenio’s marshmallow schemas handling re-instantiates schemas on each service call. Part of the fix has already been done by moving request data to ContextVar.

What's left:
- removing partial(...) nested schema workarounds
- building custom-field schemas once at startup.
- Add global schema cache

These would be the backend changes for large-record performance improvements.

Comment from #3006 
> After analyzing the pain points using the [profiler](https://github.com/inveniosoftware/invenio-app-rdm/pull/3199), the profiling helped in finding performance issues as follows:
> 
> With 5,000 authors, a single draft update takes 3.685s (84% of the time being in serialization). Key bottlenecks:
> 1. `deepcopy` operations in `make_affiliation_index` consuming 15-20% of execution time
> 2. Marshmallow schema validation/deserialization consuming 60-80% through nested `CreatorSchema`/`PersonOrOrganizationSchema`/`AffiliationRelationSchema` processing
> 3. Text sanitization
> 4. UI also renders all authors all together which is slow.
> 
> **Stack trace for de-serialization**
> Image
> 
> **Stack trace for draft validation**
> Image
> 
> **Jinja template rendering bottleneck**
> Image
> 
> **Similarly for the publish flow, 5,000 authors takes 4.158s with the pain points mentioned above.**
> 
> ### The loading of the landing page also takes ~7s bringing the total time taken starting from a saved draft with 5,000 authors being published(which again calls the update draft method for data validation before publish) and the new record landing page being displayed to a total of ~15 seconds
> 
> Subtasks:
> - [ ] Optimize heavy operations like deepcopy in backend serialization and schema validation
> - [ ] Add caching for nested schemas during serialization to speed up
> - [ ] Re-assess fields using SanitizedUnicode and alike to avoid unneeded operations
> - [ ] Optimize landing page author display (Searchable / Collapsable / Lazy loading) in template rendering
> - [ ] Optimize deposit form UI react components and validation (CreatibutorsField react component)
> 
> Please let me know if I missed something in the subtasks or if there's any doubts in the assessment! 

 _Originally posted by @sakshamarora1 in [#3006](https://github.com/inveniosoftware/invenio-app-rdm/issues/3006#issuecomment-3415126537)_

sakshamarora1:
closes: https://github.com/inveniosoftware/invenio-app-rdm/issues/3480

This PR updates the core records schema layer so marshmallow schemas can be reused safely, instead of being recreated on every service call.

- Schema instances are cached and reused via a global schema cache (get_cached_schema), so load/dump no longer build a fresh schema tree on every request.
- CustomFieldsSchema is built once at app startup through CustomFieldsSchemaRegistry, using config resolved in init_app() rather than at import time.

---

Optional (for testing) - **Currently with these optimizations it may not be needed in the end, keeping in mind the future move to marshmallow v4 should also better the performance further.**

For adding DFM (DeepFriedMarshmallow), add sitecustomize.py inside the `site/` folder of your instance
```
import re
import deepfriedmarshmallow.jit as dfm_jit
from deepfriedmarshmallow import deep_fry_marshmallow
from marshmallow import fields
from deepfriedmarshmallow.jit.plugins import register_field_inliner_factory

dfm_jit._VALID_IDENTIFIER = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_]*$")

def _constant_inliner_fixed(field_obj, context):
    if not isinstance(field_obj, fields.Constant):
        return None
    sym = f"__dfm_const_{id(field_obj)}"
    context.namespace[sym] = field_obj.constant
    return sym if context.is_serializing else None

register_field_inliner_factory(_constant_inliner_fixed)

deep_fry_marshmallow()
```

Because there is some punting done by the creator, so this is to work around it for now just for testing.

### DIFF
--- a/invenio_records_resources/proxies.py
+++ b/invenio_records_resources/proxies.py
@@ -21,3 +21,15 @@ current_notifications_registry = LocalProxy(
 current_transfer_registry = LocalProxy(
     lambda: current_app.extensions["invenio-records-resources"].transfer_registry
 )
+
+current_custom_fields_schema_registry = LocalProxy(
+    lambda: current_app.extensions[
+        "invenio-records-resources"
+    ].custom_fields_schema_registry
+)
+"""Helper proxy to get the current custom fields schema registry."""
+
+current_global_schema_registry = LocalProxy(
+    lambda: current_app.extensions["invenio-records-resources"].global_schema_registry
+)
+"""Helper proxy to get the current global schema registry."""
--- a/invenio_records_resources/registry.py
+++ b/invenio_records_resources/registry.py
@@ -3,6 +3,8 @@
 
 """Invenio-Records-Resources registry."""
 
+from .services.custom_fields.schema import CustomFieldsSchema, CustomFieldsSchemaUI
+
 
 class ServiceRegistry:
     """A simple class to register services."""
@@ -53,3 +55,36 @@ class NotificationRegistry:
         Returns an empty list if not found.
         """
         return self._handlers.get(record_type, [])
+
+
+class GlobalSchemaRegistry:
+    """A global registry for cached schemas."""
+
+    def __init__(self):
+        """Initialize the registry."""
+        self._schemas = {}
+
+    def get(self, schema_cls):
+        """Get a cached schema."""
+        if schema_cls not in self._schemas:
+            self._schemas[schema_cls] = schema_cls()
+        return self._schemas[schema_cls]
+
+
+class CustomFieldsSchemaRegistry:
+    """A registry for custom fields schemas."""
+
+    def __init__(self):
+        """Initialize the registry."""
+        self._schemas = {}
+
+    def register(self, app, fields_var, schema_cls):
+        """Register a new custom fields schema."""
+        field_property_name = schema_cls.field_property_name
+        config = app.config.get(fields_var, [])
+        self._schemas.setdefault(fields_var, {})
+        self._schemas[fields_var].setdefault(field_property_name, schema_cls(config))
+
+    def get(self, fields_var, field_property_name):
+        """Get a custom fields schema by fields variable and field property name."""
+        return self._schemas[fields_var][field_property_name]
--- a/invenio_records_resources/services/custom_fields/schema.py
+++ b/invenio_records_resources/services/custom_fields/schema.py
@@ -3,7 +3,6 @@
 
 """Custom Fields schema for InvenioRDM."""
 
-from flask import current_app
 from marshmallow import Schema
 
 
@@ -15,10 +14,9 @@ class CustomFieldsSchema(Schema):
 
     field_property_name = "field"
 
-    def __init__(self, fields_var, *args, **kwargs):
+    def __init__(self, config, *args, **kwargs):
         """constructor."""
         super().__init__(*args, **kwargs)
-        config = current_app.config.get(fields_var, [])
         self.fields = {
             field.name: getattr(field, self.field_property_name) for field in config
         }
--- a/invenio_records_resources/services/files/schema.py
+++ b/invenio_records_resources/services/files/schema.py
@@ -12,6 +12,7 @@ from typing import Mapping
 from marshmallow import RAISE, Schema, ValidationError, post_dump, pre_load
 from marshmallow.fields import UUID, Boolean, Dict, Integer, Nested, Str
 from marshmallow_oneofschema import OneOfSchema
+from marshmallow_utils.context import context_schema
 from marshmallow_utils.fields import GenMethod, Links, TZDateTime
 
 from ...proxies import current_transfer_registry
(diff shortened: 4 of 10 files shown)


invenio_records_resources/services/custom_fields/schema.py
```diff
@@ -15,10 +14,9 @@ class CustomFieldsSchema(Schema):
 
     field_property_name = "field"
 
-    def __init__(self, fields_var, *args, **kwargs):
+    def __init__(self, config, *args, **kwargs):
```
kpsherva, line 17: this is a breaking change, are you sure it is necessary?

invenio_records_resources/services/files/service.py
```diff
@@ -172,8 +172,9 @@ def init_files(self, identity, id_, data, uow=None, **kwargs):
         # if user has permission for the transfer type that is on
         # each of the uploaded files. This is done in the IfTransferType
         # permission generator.
-        schema = self.initial_file_schema.schema(many=True)
-        data = schema.load(data)
+        data, _ = self.initial_file_schema.load(
```
kpsherva, line 175:
```suggestion
        data, errors = self.initial_file_schema.load(
```
let's be explicit in naming, even if the errors will end up empty?

invenio_records_resources/proxies.py
```diff
@@ -21,3 +21,15 @@
 current_transfer_registry = LocalProxy(
     lambda: current_app.extensions["invenio-records-resources"].transfer_registry
 )
+
+current_custom_fields_schema_registry = LocalProxy(
+    lambda: current_app.extensions[
+        "invenio-records-resources"
+    ].custom_fields_schema_registry
+)
+"""Helper proxy to get the current custom fields schema registry."""
```
kpsherva, line 30: could you explain why you went for this solution?

invenio_records_resources/services/files/schema.py
```diff
@@ -103,13 +104,13 @@ class Meta:
     size = Integer(dump_only=True)
     transfer = Nested(TransferSchema, dump_only=True)
     status = GenMethod("dump_status")
-    transfer = Nested(TransferSchema, dump_only=True)
```
utnapischtim, line 106: nit: why has this been removed?

invenio_records_resources/services/files/schema.py
```diff
@@ -103,13 +104,13 @@ class Meta:
...
 
     def dump_status(self, obj):
         """Dump file status."""
+        ctx = context_schema.get()
```
utnapischtim, line 110: question: is this personal preference or does it make a difference performance wise?

invenio_records_resources/services/records/schema.py
```diff
@@ -89,7 +91,7 @@ def load(self, data, schema_args=None, context=None, raise_errors=True):
 
         token = context_schema.set(local_context)
         try:
-            valid_data = self.schema(**schema_args).load(data)
+            valid_data = self._schema.load(data, **schema_args)
```
utnapischtim, line 94: question: what could be in `schema_args`? for me it is strange, that this can be moved from constructor to method call without a problem

invenio_records_resources/services/records/schema.py
```diff
@@ -110,6 +112,15 @@ def dump(self, data, schema_args=None, context=None):
 
         token = context_schema.set(local_context)
         try:
-            return self.schema(**schema_args).dump(data)
+            return self._schema.dump(data, **schema_args)
         finally:
             context_schema.reset(token)
+
+    def schema(self, **schema_args):
```
utnapischtim, line 119: nit: what does happen with `**schema_args` here?

invenio_records_resources/services/files/service.py
utnapischtim: some linter will disagree here, because if the variable is assigned, but not used would indicate there is an error happened. so i would prefer to use `_` if the errors list is not used.

invenio_records_resources/registry.py
```diff
@@ -53,3 +55,36 @@ def get(self, record_type):
...
+    """A registry for custom fields schemas."""
+
+    def __init__(self):
+        """Initialize the registry."""
+        self._schemas = {}
+
+    def register(self, app, fields_var, schema_cls):
+        """Register a new custom fields schema."""
+        field_property_name = schema_cls.field_property_name
+        config = app.config.get(fields_var, [])
```
utnapischtim, line 84: nit: should this probably fail if `fields_var` doesn't exist? using `app.config[fields_var]` instead.

max-moser:
Closing in favor of https://github.com/inveniosoftware/invenio-records-resources/pull/692

PR #689, opened 2026-06-18
fix: report FAILED status when multipart upload no longer exists in storage

karkraeg:
## What does this PR do?

When a presigned direct-to-S3 multipart upload is abandoned (browser crash, network failure, upload timeout), the file record stays stuck in `"pending"` state forever. The user has no way to tell something went wrong, and the only fix is admin intervention.

This PR overrides `MultipartTransfer.status` to ask the storage backend whether the multipart upload is still alive. If the backend confirms it is gone, `FAILED` is returned instead of `PENDING`, surfacing the error in the UI.

**Changes:**
- `MultipartStorageExt.multipart_upload_exists()` — new method that delegates to the storage backend. Returns `True` (safe default) for backends that don't implement it, preserving current behaviour on local/non-S3 storage.
- `MultipartTransfer.status` — checks `multipart_upload_exists()`; returns `FAILED` if the upload is confirmed gone, `PENDING` otherwise. Any exception during the check is logged and falls back to `PENDING`.

## Depends on

- `invenio-s3` PR inveniosoftware/invenio-s3#49 — implements `multipart_upload_exists()` and the graceful `NoSuchUpload` handling in `multipart_abort_upload`.

## Notes for reviewers

- The `FAILED` path is only reachable for storage backends that implement `multipart_upload_exists()`. All existing backends are unaffected.
- The extra S3 `list_parts` call happens only while a file is in `PENDING` state (type `M`). After `commit_file`, the transfer type changes to `L` and `MultipartTransfer.status` is no longer called.
- All existing tests continue to pass unchanged.

### DIFF
--- a/invenio_records_resources/services/files/transfer/providers/multipart.py
+++ b/invenio_records_resources/services/files/transfer/providers/multipart.py
@@ -4,6 +4,7 @@
 
 """Multipart file transfer provider."""
 
+import logging
 from datetime import datetime, timedelta, timezone
 
 import marshmallow as ma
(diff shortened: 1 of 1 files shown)


mesemus:
Thanks for this PR. I'm a bit hesitant of calling the S3 inside the status - if I remember correctly the status is used in serialization and if I open 50 files for upload on the same record, it would get called 50 times each time user refreshes.

I'd go with https://github.com/inveniosoftware/invenio-s3/pull/49#issuecomment-4764973412 - storing the date on the transfer metadata and checking that. That would solve the expiration problem. 

Other types of problems (like S3 administrator aborting the upload directly) would not be handled by this change. We might add a background process to check the uploads and remove that - maybe writing user a notification that we have done so?

wgresshoff:
I think this sounds like a good plan! Thanks @mesemus !

PR #690, opened 2026-06-19
fix(performance): replace deepcopy with shallow copy for link expansion

max-moser:
`deepcopy()` has been identified as a costly operation, and we generally don't perform modifications on context anyway.

### DIFF
--- a/invenio_records_resources/services/base/links.py
+++ b/invenio_records_resources/services/base/links.py
@@ -6,7 +6,6 @@
 
 import operator
 import warnings
-from copy import deepcopy
 
 from flask import current_app
 from invenio_base import invenio_url_for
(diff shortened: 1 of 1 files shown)


invenio_records_resources/services/base/links.py
```diff
@@ -127,7 +126,7 @@ def vars(obj, vars):
     def expand(self, obj, context):
         """Expand the URI Template."""
         vars = {}
-        vars.update(deepcopy(context))
+        vars.update(context.copy())
```
fenekku: Replacing those 2 lines by:

```
vars = context.copy()
```

will actually be ideal. There are much fewer external links, so any impact is limited (from running tests on invenio-rdm-records it looks good).

fenekku (approved):
Thanks. Just further cleanup/optimization.

PR #692, opened 2026-06-23
Reusable marshmallow schema instances in services

max-moser:
PR based on @sakshamarora1 's PR https://github.com/inveniosoftware/invenio-records-resources/pull/687 but without the schema registry (pulled the schema initialization into the service constructor instead).

This PR targets `maint-10.x` because I'm trying stuff out with a v14 setup.

### DIFF
--- a/invenio_records_resources/services/files/components/metadata.py
+++ b/invenio_records_resources/services/files/components/metadata.py
@@ -1,5 +1,6 @@
-# SPDX-FileCopyrightText: 2021-2024 CERN.
+# SPDX-FileCopyrightText: 2021-2026 CERN.
 # SPDX-FileCopyrightText: 2025 CESNET.
+# SPDX-FileCopyrightText: 2026 TU Wien.
 # SPDX-License-Identifier: MIT
 
 """Files metadata component components."""
@@ -44,8 +45,7 @@ class FileMetadataComponent(FileServiceComponent):
 
     def update_file_metadata(self, identity, id_, file_key, record, data):
         """Update file metadata handler."""
-        schema = self.service.file_schema.schema(many=False)
-        validated_data = schema.load(data)
+        validated_data, _errors = self.service.file_schema.load(data, raise_errors=True)
         record.files.update(file_key, data=validated_data)
 
     def update_transfer_metadata(
--- a/invenio_records_resources/services/files/schema.py
+++ b/invenio_records_resources/services/files/schema.py
@@ -1,7 +1,8 @@
-# SPDX-FileCopyrightText: 2020-2024 CERN.
+# SPDX-FileCopyrightText: 2020-2026 CERN.
 # SPDX-FileCopyrightText: 2020 European Union.
 # SPDX-FileCopyrightText: 2025 CESNET.
 # SPDX-FileCopyrightText: 2025 Graz University of Technology.
+# SPDX-FileCopyrightText: 2026 TU Wien.
 # SPDX-License-Identifier: MIT
 
 """File schema."""
@@ -12,6 +13,7 @@ from typing import Mapping
 from marshmallow import RAISE, Schema, ValidationError, post_dump, pre_load
 from marshmallow.fields import UUID, Boolean, Dict, Integer, Nested, Str
 from marshmallow_oneofschema import OneOfSchema
+from marshmallow_utils.context import context_schema
 from marshmallow_utils.fields import GenMethod, Links, TZDateTime
 
 from ...proxies import current_transfer_registry
@@ -25,7 +27,7 @@ class BaseTransferSchema(Schema):
     """
 
     type_ = Str(attribute="type", data_key="type", required=True)
-    """Transfer type. Required field, the initial transfer type is filled 
+    """Transfer type. Required field, the initial transfer type is filled
     automatically by the InitFileSchema."""
 
     class Meta:
@@ -103,13 +105,13 @@ class FileSchema(Schema):
     size = Integer(dump_only=True)
     transfer = Nested(TransferSchema, dump_only=True)
     status = GenMethod("dump_status")
-    transfer = Nested(TransferSchema, dump_only=True)
 
     def dump_status(self, obj):
         """Dump file status."""
+        ctx = context_schema.get()
         transfer = current_transfer_registry.get_transfer(
             file_record=obj,
-            file_service=self.context.get("service"),
+            file_service=ctx.get("service"),
             record=obj.record,
         )
         return transfer.status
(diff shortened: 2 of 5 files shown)


invenio_records_resources/services/records/schema.py
```diff
@@ -110,6 +112,6 @@ def dump(self, data, schema_args=None, context=None):
 
         token = context_schema.set(local_context)
         try:
-            return self.schema(**schema_args).dump(data)
+            return self.schema.dump(data, **schema_args)
```
Samk13, line 114: Question: what should happen if a caller passes [schema_args](https://marshmallow.readthedocs.io/en/stable/marshmallow.schema.html#schema) like `dump_only` or `exclude` as [dump](https://marshmallow.readthedocs.io/en/stable/marshmallow.schema.html#marshmallow.schema.Schema.dump) accept `many` param only? 
These are no longer passed to the schema constructor.

invenio_records_resources/services/records/schema.py
```diff
@@ -1,10 +1,11 @@
-# SPDX-FileCopyrightText: 2020-2024 CERN.
-# SPDX-FileCopyrightText: 2022 TU Wien.
+# SPDX-FileCopyrightText: 2020-2026 CERN.
+# SPDX-FileCopyrightText: 2022-2026 TU Wien.
 # SPDX-FileCopyrightText: 2025 Graz University of Technology.
 # SPDX-License-Identifier: MIT
 
 """Record schema."""
 
+import warnings
```
Samk13: Looks like it's a leftover?

invenio_records_resources/services/records/schema.py
sakshamarora1: Fair point, but to me seems like an anti-pattern to keep on having that abstraction. It accepted constructor arguments during a .dump() or .load() call even when those parameters did nothing for that specific operation, like the example you mentioned. What do you think?

sakshamarora1: But yes, it would now throw an Error from Marshmallow's dump instead of not doing anything.

invenio_records_resources/services/files/schema.py
```diff
@@ -103,13 +105,13 @@ class Meta:
     size = Integer(dump_only=True)
     transfer = Nested(TransferSchema, dump_only=True)
     status = GenMethod("dump_status")
-    transfer = Nested(TransferSchema, dump_only=True)
 
     def dump_status(self, obj):
         """Dump file status."""
+        ctx = context_schema.get()
```
utnapischtim, line 111: question: is this change a personal preference or does it make a difference performance wise?

invenio_records_resources/services/files/components/metadata.py
```diff
@@ -44,8 +45,7 @@ def init_files(self, identity, id_, record, data):
 
     def update_file_metadata(self, identity, id_, file_key, record, data):
         """Update file metadata handler."""
-        schema = self.service.file_schema.schema(many=False)
-        validated_data = schema.load(data)
+        validated_data, errors = self.service.file_schema.load(data, raise_errors=True)
```
utnapischtim: nit: should we replace `errors` with `_` to make linters happy? regarding `unused variables`?

invenio_records_resources/services/records/schema.py
```diff
@@ -89,7 +90,7 @@ def load(self, data, schema_args=None, context=None, raise_errors=True):
 
         token = context_schema.set(local_context)
         try:
-            valid_data = self.schema(**schema_args).load(data)
+            valid_data = self.schema.load(data, **schema_args)
```
utnapischtim, line 93: question: how do we know, that this every time works? it feels strange, that parameters which were given before to the constructor can be given to the method too without any consequences.

invenio_records_resources/services/records/service.py
```diff
@@ -65,14 +66,22 @@ def record_to_index(self, record):
 class RecordService(Service, RecordIndexerMixin):
     """Record Service."""
 
+    def __init__(self, config):
+        """Constructor."""
+        super().__init__(config)
+        if schema := getattr(config, "schema", None):
```
utnapischtim, line 72: oh, first walrus operator that i see in invenio!

utnapischtim, line 72: nit: the original implementation didn't check if self.config.schema exists, why do we do it now here?

invenio_records_resources/services/files/schema.py
```diff
@@ -103,13 +105,13 @@ class Meta:
     size = Integer(dump_only=True)
     transfer = Nested(TransferSchema, dump_only=True)
     status = GenMethod("dump_status")
-    transfer = Nested(TransferSchema, dump_only=True)
```
utnapischtim, line 106: nit: why has this been removed?

max-moser: because it's a duplicate from two lines above

invenio_records_resources/services/records/service.py
max-moser: there's a couple more, at least in `rdm-records`

max-moser: backwards-compatibility

i vaguely remember running into a scenario where at least one service config did not have a schema field set, which caused issues
this left two options: either making sure that *every* record service config has a schema field (which includes services/configs in other packages, like invenio-requests), or adding graceful handling in the constructor here

though admittedly i haven't reevaluated after adding this guard

invenio_records_resources/services/files/schema.py
max-moser: it does not make a performance difference

invenio_records_resources/services/records/schema.py
sakshamarora1: [Marshmallow allows it](https://marshmallow.readthedocs.io/en/stable/top_level.html#marshmallow.Schema.load)
if you see the `initial_file_schema` property, it is used to validate a list in init_files, but the underlying file_schema is also used elsewhere for single file metadata update/dump. Keeping the args in the method keeps the cached schema instance stateless and reusable. The arguments override the instaatiated schema object's attributes so it's the same thing. It is required for the global schema caching.

invenio_records_resources/services/files/schema.py
utnapischtim: ahhh, thanks, didn't saw that

invenio_records_resources/services/records/service.py
utnapischtim: so in the end it is something like better safe then sorry for the production instance

ptamarit:
See linked PR #724 where these changes are merged to master.

